PR #33, opened 2017-12-03
Enable default selection

yymao:
This PR enables the "default selection" feature. Catalogs or tests that has set `included_by_default` in their config file will be run if no `-c` or `-t` options are provided.


evevkovacs:
Nice!. You just did one of the projects on my list for sprint week.

yymao:
Note that this PR takes descqa to 2.4.2 and it now requires GCRCatalogs 0.6.5

PR #37, opened 2017-12-05
Updated basic_test to include value ranges

evevkovacs:
Changed basic_test.py to include ranges of values in output
Added yaml file to select calc_min_max option


evevkovacs:
branch name is not indicative of this PR

PR #39, opened 2017-12-07
Nvsz refactor complete

evevkovacs:
Still need to refactor the catalog reading to be more efficient


yymao:
Thanks, @evevkovacs. Let's try to reorder the two loops tomorrow morning!

Issue #42, opened 2017-12-08
2pt correlation (e.g. ellipticity-direction) for IA model testing


yymao:
@EiffL, @patricialarsen, and @jablazek have been working on getting 2pt correlation (e.g. ellipticity-direction) for IA model testing. 

Can one of you list the items that we should test? And we can see if we need separate one issue for each of them. 

Some techniques used here are related to #10 #35 but the purpose and validation datasets would be different. 

P.S. @patricialarsen I cannot assign you. Please register your GitHub account to the DESC roster.

yymao:
So far we only have a theoretical prediction on w_g+ from @jablazek (plot taken from [this run](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-08_34&test=IA_theory)) 
![image](https://user-images.githubusercontent.com/3792659/34002625-13f4d3de-e0c1-11e7-94fd-4e3d5f8f0a86.png)

@EiffL @patricialarsen any updates on implementing w_g+ (or other 2pt statistics) from mock data?

patricialarsen:
@yymao I have the shear-shear tests implemented (in real space) in my version of this (theory and data). This seems to work fine, and I am now expanding this to kappa-kappa tests etc. and adding in criteria for passing the test.

yymao:
@patricialarsen that sounds great! Can you share some plots with us in this issue? If you can do a PR of whatever you have coded up that would be even better! (We do encourage early PR!)

patricialarsen:
@yymao, yes I'll sort that all out once I have more reliable internet. I'm in Mexico at the moment and will spend the weekend on airplanes so this week isn't ideal! By the way this 2-pt testing is as much for validation of the weak lensing as for IAs.

rmandelb:
A note about that theory prediction from @jablazek - that must be for the 2-halo term only.  If we're comparing the mocks against that, then we'd need to be careful what scales we compare on.

rmandelb:
@patricialarsen - I wanted to touch base with you about this validation test.  A few questions:

1. You mentioned that this test is for lensing as well as for IAs.  If so, I think it would be worth splitting into two tests because they are testing rather different things for which our validation criteria will differ quite a bit.  Do you agree, and if so, could you think about the proper split to make and open a new issue/issues as appropriate?

2. I know that @chihway had done some tests of shear-shear correlations already, and I wanted to know if you have seen that work and might consider incorporating some of it? (or whether there was some obstacle to doing that)

[...]

PR #45, opened 2017-12-08
adding first draft of IA theory module. Still needs work.

jablazek:
This won't work yet. We still need to get FAST-PT in the descqa environment.


jablazek:
FAST-PT code is available here: https://github.com/JoeMcEwen/FAST-PT

be sure to use the latest on master rather than the latest tagged version.

yymao:
@jablazek check this out https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-08_24&test=IA_theory

yymao:
While this is still an early prototype, it will eventually close (a portion of) #42

Issue #7, opened 2017-11-06
dN/dmag test


yymao:
See https://github.com/LSSTDESC/DC2_Repo/issues/21 for detail and validation data from HSC.

- [x] code to reduce mock data
- [x] code that works within DESCQA framework
- [x] validation data
- [ ] validation criteria

duncandc:
I am going to start working on this.

yymao:
Here's a preliminary result from this test, made by @duncandc: (and the [corresponding entry on DESCQA web interface](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-07_26&test=ApparentMagFuncTest))
![image](https://user-images.githubusercontent.com/3792659/34000936-b0a0a690-e0bc-11e7-931b-bd644db1930e.png)

Lines are protoDC2 and buzzard_test, points are HSC. Difference in normalization may be due to a bug (still be explored). 

See also https://github.com/LSSTDESC/DC2_Repo/issues/21 for related comments.

rmandelb:
@yymao @duncandc - There has been further discussion of this test in #50 (see specifically from https://github.com/LSSTDESC/descqa/issues/50#issuecomment-360517640 onwards).  A few outstanding questions:

- as discussed there, we'd like to have some way of disabling this test for catalogs that have redshift cuts such that the N(<mag) calculation is significantly affected.  Can this be implemented based on the maximum redshift of the light cone?
- the last status of this issue was that there were bugs to work out.  Is this still in progress?
- validation criterion: we had discussed this on https://github.com/LSSTDESC/DC2_Repo/issues/21 and the proposal I would like to make is that at limiting magnitudes of 23, 24, and 25 in i-band, we should require 20% agreement in N(<mag).  At limiting magnitudes of 26 and 27, we should require 40% agreement.  Pinging a few people who may have comments on this: @janewman-pitt-edu , @msimet .

yymao:
@rmandelb To answer your questions --- 
- Yes, that can be easily implemented.
- We have not made much progress since the Sprint Week (but @duncandc can correct me if that's not the case). I am not sure if there was really a bug in getting dN/dmag from the mocks, or just that the way the validation data is constructed is not consistent with how we do it with the mock.
- Those criteria can be implemented. As mentioned above, we need to make sure the validation data is correctly/consistently constructed.

janewman-pitt-edu:
The criteria Rachel suggests sounds reasonable to me.

[...]

PR #47, opened 2017-12-08
apparent magnitude function test

duncandc:
This is a premature pull request for the apparent magnitude function test.


yymao:
This is a prototype for the test requested in #7

yymao:
Output of the current version of this validation test can be viewed at https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-07_26&test=ApparentMagFuncTest

Below I copy and paste @rmandelb's comments from https://github.com/LSSTDESC/DC2_Repo/issues/21

- [ ] the plot should say which is validation data and which is the sim
- [ ] the summary plot should say which sim is which
- [ ] should have an axis label for the horizontal axis that says what band is being shown
- [ ] it would be helpful to display number densities in number per square arcmin, rather than number per square degree (I know it's a minor point, but many observers have a rough rule of thumb for expected density per square arcmin in imaging data, so looking at the plot in those units without having to divide by 3600 makes it easier to do sanity checks).
- [ ] having the vertical axis explicitly say this is a number density rather than a raw number -- e.g., by noting the units -- would be helpful.

yymao:
Further suggestions on the validation criteria and plot enhancements can be found at https://github.com/LSSTDESC/DC2_Repo/issues/21

yymao:
@duncandc @rmandelb what's the current status of this test --- is it ready for review? If not, what are the necessary changes to be made?

rmandelb:
@yymao - I do not know the status; I have not seen an update for a long time. @duncandc , is there a chance this can be updated during this week?  We would like to have the dN/dmag test in place.

evevkovacs:
@rmandelb Is the level of agreement between catalog and data acceptable?
@duncandc @yymao For my information, what exactly does descqagen 2018 do? Thanks

rmandelb:
@yymao am I reading the code correctly that it's requiring a 20%-level match for all mags between 23-28.5? (i.e., the default limits are 24-27.5 but then it goes 1 mag fainter than the faint limit and 1 mag brighter than the bright limit)  This is indeed more strict than we had intended.  Our proposal as I recall had been more like a 20%-level match for 23<i<25.3 (within ~1 mag of the gold sample limit) and 40%-level match for 25.3<i<27.  The rationale behind this is that 23<i<25.3 objects will dominate the lensing constraints, so we want to match their density reasonably well, but the ones from 25.3<i<27 are mostly important because of the resulting blending with low-level detections (and this should be reasonable but it doesn't have to be as good a match as i<25.3, especially since the validation dataset is itself being extrapolated at this point).  And no validation criterion should be imposed beyond i=27 because they just don't matter that much since it's >1.5 mags beyond the nominal limit of the sample we'll use.  Does that sound like what you remember @msimet ?

Also, I highly recommend just imposing the criterion in the i band, and relying on the color tests for getting the other bands right.  It's still nice to have the plots in the other bands, I just don't think we need to cut on them.

msimet:
Yeah, that sounds about right--looks like the initial proposal was [here](https://github.com/LSSTDESC/descqa/issues/7#issuecomment-361585845), concurred by Jeff.  It still sounds reasonable to me.

katrinheitmann:
So in that issue there is also the question by Rachel about flattening 
out. How reliable is the extrapolation of HSC?

Thanks!

aphearin:
The plot below may be useful to help get a physical intuition for which population(s) of galaxies are under scrutiny in this particular test. 

hsc_apparent_mag_to_mstar_vs_z

aphearin:
@rmandelb - re: deblending. Yes, I understand and agree that's important and the primary driver here. Part of the reason I made this plot was to illustrate that, for most redshifts of interest (not z>~1), the relevant galaxies are so faint that they could just be sprinkled in randomly.

aphearin:
Are we confident that the power law extrapolation is appropriate? @rmandelb expressed some skepticism [in this comment ](https://github.com/LSSTDESC/descqa/issues/7#issuecomment-364329846)about the flattening of dn/dmag for i > 26, but this was never resolved in the subsequent discussion.

aphearin:
@yymao - before merging the PR of this HSC dN/dmag test, could you make sure that this test implements the feature where the plotted data are provided in ASCII form? This is a super nice feature of many DESCQA tests that makes it a lot quicker to improve the catalog in response to the test.

yymao:
@aphearin Good point. Will do. (cc @duncandc)

evevkovacs:
@yymao @duncandc It is also really helpful to print out any selection cuts in the test files. I usually put them in the header. See for example the text files for the N(z) test. Thanks.

evevkovacs:
@duncandc  @yymao Katrin suggested (and also Rachel some time ago) that it would be very helpful to have a panel in the validation plot showing the ratio of the catalog to the HSC extrapolation. I am happy to help implement this if you need manpower to do it.

katrinheitmann:
It would be also good to indicate where the extrapolation starts. Thanks!

rmandelb:
@aphearin 

> Are we confident that the power law extrapolation is appropriate?

Generically, very deep surveys find power law slopes of the number counts.  From a quick search, you can see this happening down to 28th magnitude in z-band in the Hubble ultra deep field in figure 6 of https://arxiv.org/pdf/astro-ph/0607632.pdf.  (The curve starts to flatten out beyond that, but they also are starting to be incomplete at that point.)  There are other examples of this in other optical bands as well, albeit relatively few that go that deep because, well, most places haven't been observed that deep.  But very deep observations like those in the UDF provide the general justification for saying "we think it's fair to extrapolate the number counts like a power law out to ~28th magnitude".  The extrapolation is plotted beyond that, which is possibly not wise, but we aren't using it for validation so I think it's not causing any actual problems.

With that said, given that the power-law fit parameters always have some uncertainty associated with them, that's another reason to weaken the validation criteria (as I've suggested) in the regime of extrapolation.

I agree with Katrin's suggestion that it is a good idea to indicate the regime of extrapolation, even though there is very good justification to believe that a power-law extrapolation is fair.

@evevkovacs 
> I aslo have a question about the HSC extrapolation. This plot shows the data turning over above magnitude 26.  Presumably, the extrapolation is not really trustworthy above this value.

Hopefully my explanation to Andrew has made you more confident in the validity of a power-law extrapolation, at least out to ~28th magnitude (we don't use it for anything beyond that, but I think it would be justifiable to stop plotting it after that point).

rmandelb:
One other thing @evevkovacs : the HSC completeness as a function of magnitude has been estimated and we know it's incomplete in the regime of flattening, so that's actually a reason to believe the power-law extrapolation *more* than we believe the data that shows a flattening.

rmandelb:
I confess I got curious about other measurements of number counts, especially since what I showed above was z-band (rather than i-band, which is what I proposed we use for this test).  So I did a search and then realized there's a nice summary figure in the HSC data release paper: see figure 16 of https://arxiv.org/pdf/1702.08449.pdf

This shows a comparison of number counts for HSC ultradeep (which is deeper than the HSC deep sample we've used here) vs. other very deep observations, all for i-band.  I believe this figure justifies the use of a power-law number counts specifically in i-band to around i=27, with some uncertainty after that point.  So that's not quite as deep as I was claiming in my comment above (based on the z-band UDF results I had suggested 28th magnitude).  However, since I proposed earlier that we only impose validation criteria against the extrapolated number counts up to a limit of 27th magnitude in i-band, I think it's all good?  If people would prefer not to show the extrapolation from HSC beyond that point, I think that would be a reasonable change.

aphearin:
@yymao - what effect are you thinking of that makes the lightcone mock incomplete?

yymao:
@aphearin I was thinking that the mock should be complete in absolute magnitude and hence incomplete in apparent magnitude. Does that make sense?

duncandc:
@yymao Presumably isn't that one of the effects this test is supposed to address? (sorry to jump in just now, I have been traveling...)

yymao:
@duncandc maybe? I am actually not sure. So there are two different potential issues in the mock: (1) the hard redshift cut at z=1 and (2) sim/model resolution limit. After looking at @aphearin's plot above and some offline discussion with @aphearin, we think issue (2) is probably ok (i.e., do not have huge impact on this test). Issue (1) is probably still problematic, but I don't know how to mitigate it for protoDC2. 

But in any case, I think we all agree that the current criterion is too stringent and should loosen it. And it would be visually helpful to indicate some of the assumptions that goes into the plot (e.g. where the data is extrapolated). 

@duncandc can you update the test accordingly (see all the comments above), or do you have different opinions/suggestions?

duncandc:
I can definitely update the criteria for the test and indicate both the extrapolation region and even the  "data" used.

rmandelb:
Right - I forgot about the z<1 issue.  That could indeed be driving the flattening at fainter mags, though there are plenty of z>1 galaxies even at the magnitudes below 25 where the mocks are already above the data.  So once we have cosmoDC2 (which I thought was going to z=2?) the mocks will be even further above the data.

aphearin:
@rmandelb @yymao @janewman-pitt-edu - In the attached plot, I wanted to know the impact of two different things on the dn/dmag HSC test:

1. Are we being impacted by the z < 1 cut that is present in the mock but not present in the data?
2. Are we being impacted by the M*>10^6 cut that is present in the mock but not present in the data?

A preliminary look at the first question appears on the left panel, the second question on the right panel. 

hsc_dndmag_redshift_mstar_cut_systematics

## Impact of the redshift cut
In the left panel, we can see that there are two basic effects of imposing redshift cuts:
* As the redshift cut is applied more aggressively, the overall normalization decreases.
* As the redshift cut is relaxed, the turnover in the LF happens at fainter magnitudes. 

This means that if we used the same model that is tuned on AlphaQ, and just scaled it up to cosmoDC2, we would improve the shortcoming of protoDC2 at faint magnitudes (i >~ 26) since the turnover would get pushed to fainter mags, but we would also increase the overall normalization, which will worsen the agreement at brighter magnitudes (i <~ 26). 

## Impact of stellar mass incompleteness
In the right panel, we can see that there is really only one effect of the stellar mass resolution effect: as stellar mass incompleteness improves, the slope turns over at fainter magnitudes. 

I noticed that Buzzard appears to have the same turnover in this test, as seen in [this DESCQA plot](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-12_42&test=ApparentMagFuncTest_HSCr&right=2018-04/2018-04-12_42/ApparentMagFuncTest_HSCr/summary.png), so I thought I would tag @j-dr in case he has insight into other contributing factors. In the Buzzard plot, please ignore the protoDC2 curve that appears there, which is out of date - the relevant protoDC2 catalog is v4.5, whose DESCQA test has been given earlier in this thread and [is repeated here for convenience](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-05-23_12&test=ApparentMagFuncTest_HSCi&catalog=proto-dc2_v4.5_rescale).

janewman-pitt-edu:
Very interesting!!! Nice idea!

yymao:
Is it possible that the stellar mass-luminosity relation is off?

aphearin:
@rmandelb - in the model, the `<M* | Mpeak>` relation is entirely independently modeled from the `<Mr | M*>` relation, so @yymao is almost certainly correct that this latter relation needs modification. I am actually working on this today. My plan is to decrease the normalization of this relation so that the `i<25` match with HSC is improved. Based on further experimentation since my last post, I actually think that stellar mass incompleteness in the UniverseMachine below M*<10^7 is actually the culprit for the incorrect turnover in the power law for i>26, so I will also be working on fixing this so that we improve the faint-end. 

As for the massive end, that is still quite puzzling to me. I really do not know how to reconcile the CMASS test that just came in with the Moustakas+13 SMF test. The SMF test is confirmed by the UniverseMachine code, which uses Moustakas+13 as fitting data, so two fully independent implementations of this test produce consistent results. Off the cuff, I don't see how color modeling can account for a 6x factor, but I have not experimented with this so I don't know for sure. I could certainly try modifying the `<Mr | M*>` relation at high mass and intermediate redshift, but this will sensitively impact all RedMapper tests, which currently pass across the board; given how much labor was involved in getting those right, this makes me pause before modifying this regime. I will take a look at how much wiggle room there is in this relation and report on the results, but in the meantime, other ideas are welcome.

yymao:
Here's the stellar mass-Magnitude relation in v2.1.2 and v4.5 for reference. Is the redshift-evolution in v4.5 too big?

image

rmandelb:
So from z=0 to z=1, the v4_5 model has a factor of a few evolution in <M* | L>?  That seems steep.  There are all the usual caveats in stellar mass estimates from data but it seems to me that some empirical comparison would be worthwhile, e.g. in COSMOS (which has the nice 30-band photometry to work from).  I could probably dig up a <M* | L> in redshift bins from there if it would be useful?  The catalogs are all sitting around on my computer somewhere...

aphearin:
Thanks @yymao - you have anticipated one of the very first things I will be experimenting with. In the model, we have explicit parametric control over the redshift-dependence of the `<Mr | M*>` relation, which is what I was going to experiment with. The reason for the increased redshift evolution was so that we passed the DEEP2 dn/dz test; this parametric leverage is also exploited to pass the HSC test. In principle, the redshift evolution can itself be M*-dependent, which would give more freedom to pass these tests in addition to the CMASS test. However, implementing and calibrating that will require significantly more labor than retuning existing parameters. I'm certainly willing to put in that labor since CMASS is a very important anchor of the high-mass end at intermediate redshift, and it would certainly be nice if cosmoDC2 is useful for accurate science in this regime. But this will take some time and implementing these improvements will necessarily come with a considerable delay of a week or perhaps even more. 

In order to do this effectively, it would be enormously helpful to have the code @Andromedanita wrote to select the CMASS sample.

rmandelb:
Also, @aphearin - if you are tuning to match dN/dmag from HSC, the z<1 cut in the sims is still relevant.  Since the median redshift is ~0.9 for an i<2.53 sample, it seems like you should actually be aiming for about 1/2 the observed cumulative counts evaluated at that limiting magnitude?

Regarding the CMASS test, one thing to keep in mind is that that has not been reviewed (I don't think?) so it might be worth an independent check?  With that said, if we think the stellar masses are ~right and the luminosity needs adjusting, then the CMASS test will be quite affected because of the mag cut.  So it seems to me that if you plan to keep the stellar mass distribution (matched to Mpeak) fixed and change the stellar mass-to-light ratio, then we should simply see what happens with the CMASS test after that point and not worry about it right onw?  It seems the sign of the change will be correct, if you're reducing the number of galaxies at fixed mag then the CMASS-like sample size should drop.

aphearin:
> Since the median redshift is ~0.9 for an i<2.53 sample, it seems like you should actually be aiming for about 1/2 the observed cumulative counts evaluated at that limiting magnitude?

Yes, understood and agreed.

> So it seems to me that if you plan to keep the stellar mass distribution (matched to Mpeak) fixed and change the stellar mass-to-light ratio, then we should simply see what happens with the CMASS test after that point and not worry about it right now?

Ok, I will do my best to strike a balance between wall clock time and mock accuracy. All this input is super, super helpful, so thanks for all the engagement @rmandelb @yymao @Andromedanita @katrinheitmann.

j-dr:
Sorry to be late to the game, but I just wanted to confirm that the issue in Buzzard is indeed that the current catalogs aren't complete to the flux limit of LSST/HSC which is why the turnover in n(<m) occurs. There aren't any additional concerns that @aphearin hasn't already pointed out.

aphearin:
Thanks for chiming in @j-dr - that makes me even more optimistic that adding in synthetic galaxies with very low stellar mass will help improve the issue. That gives an independent handle on i<~25 normalization and i>26 slope.

rmandelb:
@aphearin - to satisfy my own curiosity, I took the COSMOS 30-band photo-z catalog down to i=25, and found the median stellar mass in bins of luminosity for z<0.4 and 0.7<z<1.  The result looks like this:
image
(after restricting to a range of luminosity that is present in both redshift ranges; of course the low redshift one actually goes much fainter)

The difference between the two, in Delta(log10(stellar mass)), is here:
image
(that's just the direct subtraction of the two curves in the top plot)

aphearin:
@rmandelb - thanks for sharing, this is a helpful basis of comparison. Could I ask a favor? Would you mind making the same plot but with independent and dependent variables reversed? In the DC2 model, M* is the independent variable. So knowing delta_mag as a function of M* is the most directly portable knowledge.

janewman-pitt-edu:
@rmandelb : As I recall the Laigle catalog looks good at 0.4 < z < 1.5, but shows some oddities at lower redshift.  Could you maybe do the analysis comparing z=0.4-0.6 to a higher-z bin?

erykoff:
The redMaPPer tests do pass but the clusters are a bit richer than in DES. Not x6 but consistent sign with the CMASS increase.

aphearin:
@erykoff - could you fill in a few details about your comment on your mass-richness test? The reason I ask is that the mass range of CMASS is quite large - see [this plot](https://github.com/LSSTDESC/descqa/issues/109#issuecomment-391516110) made by @Andromedanita, for example. I would think that galaxies <L* would dominate the richness, but I don't know for sure what you mean by "richness" when you use it here. Does this refer to red-sequence-richness? Or does this richness refer to all-galaxy membership? In the mild excess you are reporting, does it pertain to galaxies in the CMASS mass range? Or is it more closely connected to lower mass galaxies?

erykoff:
Sorry to comment and run. I’m back from Paris now. Anyway, Joe has the plot of the mass richness relations which show that the halos/clusters have more “red galaxies brighter than 0.2L*” at fixed mass in 4.5 vs DES. (When I say richness you can assume I mean redMaPPer richness). This is indeed a different mass scale than cmass. But it also implies more red galaxies in the CMASS redshift range. Assuming the correct-ish luminosity function, though, increased richness means increased numbers at all luminosities, which includes the CMASS selection. But I haven’t checked any of this, this was just my quick thought based on seeing the comment thread.

rmandelb:
@aphearin @janewman-pitt-edu - I addressed both of your comments.  There are now 4 redshift ranges, and I reversed the dependent vs. independent variables before doing the calculations.  The first plot legend gives the exact redshift ranges and also my names for them that are used in the 2nd plot.
image
image

Any other changes are easily made, so just let me know.

aphearin:
Many thanks @rmandelb - this is a very helpful guideline for building a model for the redshift evolution that has sufficient flexibility. I've been chipping away at this; I'll have a progress update by Monday or Tuesday.

aphearin:
Ok @rmandelb - since you were nice enough to offer to make small variations, can I ask to look at the bottom plot, but instead always looking at bin_N - bin_1? That's the way the DC2 model is formulated, since this allows me to fine-tune the SDSS low-redshift anchor and then only modify changes relative to that anchor.

rmandelb:
image

janewman-pitt-edu:
So the CMASS test is based on applying the same color cuts used for the SDSS sample?  I certainly would think you could change density by 6x given that we _know_ the colors are off, and just a bit of change in color or brightness will yield a big change in density when you're talking about things that are >L*.  I would never want to give a test based on the CMASS color cut much weight unless (at minimum) you are shifting all the colors to make sure the red sequence color vs. z matches that used to design CMASS.

janewman-pitt-edu:
Basically: if you're mixing a color test in with trying to match CMASS clustering, that's a bad idea.  Check color distributions separately (which DESCQA2 should do).  But when you fail the color test, PLEASE don't try to apply clustering/density tests that assume your colors are right and tune anything to them!!!

janewman-pitt-edu:
Also: I would **STRONGLY** advise against trusting bin 1 in this dataset too much, based on my experience with it.  Compare to SDSS, not Laigle et al., at that redshift range.

aphearin:
Just posting a brief update on the status of this issue. In order for the redshift evolution of the luminosity function to exhibit the character of the M*-dependence shown in the plot shared by @rmandelb, it was necessary to implement a new model for this evolution. The first plot demonstrates the new model. The vertical axis in the left and right panels shows the difference in restframe absolute r-band magnitude relative to z=0. 
 
sigmoid_redshift_evolution_model_v4p6

The next plot below shows how the new model fares in the HSC dn/dmag test. In the right-hand panel below, I additionally show the fractional error. The gray band highlights the faint-end validation regime, and the dashed horizontal black lines guide the eye to the 40% error tolerance that @rmandelb and @janewman-pitt-edu suggested was a reasonable target for DC2.

hsc_dndmag_v4p5_vs_v4p6

Re: timeline. There are still many things to check before taking this new calibration to the production of v4.6. Thus far I have only had time to make sure the DEEP2 dn/dz test looks good, and that the baseline z=0 SDSS luminosity function is still in ~10% agreement. I expect it will take another ~2 days to thoroughly test the new calibration and integrate the changes into the pipeline, and it takes ~24 hours to produce a full-fledged new mock, so the earliest `v4.6` could be ready is this Friday. 

Having spent some more time studying this problem, I think this validation test result could potentially change considerably when switching to a z < 3 lightcone for cosmoDC2, but it is difficult to gauge how much. After `v4.6`, we will likely need to perform another calibration of the model parameters on a z<3 cutout. That will (hopefully) not be as time-consuming as the current effort to produce `v4.6` because the current effort has involved implementing much more flexible functional forms for the model. 

CC @yymao @katrinheitmann @evevkovacs @dkorytov

yymao:
@aphearin this looks great! Thanks for the update! 

@duncandc With v4.6 coming up soon, do you have some time to make the changes to this test as we discussed above? I'd like to merge this as soon as possible. Thanks!

aphearin:
### Upshot 
I'm proposing we move forward to production with v4.6ab shown in the plots of this update, and postponing further calibration until we switch to the 0 < z < 3 Outer Rim cutouts, which anyway will be very soon. Comments welcome. 

### New dn/dmag calibration

Since the protoDC2 v4.6 lightcone will still be limited to 0 < z < 1, I have been easing on the side of under-predicting dn/dmag, as discussed with @rmandelb. This first plot shows the current status with the dn/dmag test, with two alternative (very similar) models. 

hsc_dndmag_v4p6_calibration

Note that this is not quite as close of an agreement as shown in the previous post, because I think it will be better to hold off on calibrating the addition of new synthetic low-mass galaxies until we switch to the high-z lightcone. 

### New dn/dz calibration

I would actually prefer to have the model in the HSC plot with smaller values of dn/dmag(r~23), because of the boost I expect we'll get from higher redshift contributions. But it's a little tough to find a region of model parameter space that also passes the DEEP2 dn/dz validation criteria for bright galaxies at z>~0.75. You can see that in the upper left panel of this plot:

deep2_dndz_v4p5_vs_v4p6

The current v4.6ab are as far as I've gotten with striking the balance between these two measurements. 

The new calibrations also have much more realistic cluster BCG CLFs at high redshift, which were grossly bright in v4.5, as shown in this plot. 

group_cluster_bcg_clf_v4p6_calibration

These extremely bright centrals were an unintended consequence of calibrating v4.5 to the problem region z>0.75 DEEP2 dn/dz(r<22). I had been locally checking the z=0 CLF in my own validations, but not the high-z CLF. I have now locally implemented a regression test to catch this in the future, but this also was noticed by @chto , so this validation failure will eventually get caught by DESCQA once PR #102 gets merged. 

The reason I think we should move forward with more or less the currently shown v4.6 is that we'll need to recalibrate again on the deeper lightcones anyway, and I think this upgrade at least broadly improves upon the v4.5 calibration of the luminosity function evolution. Comments welcome, cc @janewman-pitt-edu @erykoff @j-dr @katrinheitmann @evevkovacs @dkorytov

janewman-pitt-edu:
Your proposed plan seems reasonable.  I would worry more about matching r<23 / r<24 than r<22 since numbers are a bit sparse there.

erykoff:
At risk of asking a dumb question ... do we have errors on the DEEP2 numbers?

janewman-pitt-edu:
I’ve never really tried — the errors would ultimately be driven by sample/cosmic variance and so aren’t trivial to quantify.  Basically the data can be consistent with a form propto $z^2 e^{-z/z0}$ or $z^2 e^{-{z/z0}^1.5}$ 

Certainly any formal fit errors are small.  You can see from the attached plot that the z0 vs. r relation is pretty tight.

[rmaglim.pdf](https://github.com/LSSTDESC/descqa/files/2058377/rmaglim.pdf)

rmandelb:
@aphearin - your proposed plan also sounds good to me.  Is there a need to re-check some of the other tests for v4.6?  (for example: are sizes determined by a size vs. stellar mass relation, or size vs. luminosity relation?  If the latter, then the size-lum test should be the same as before.  If the former, then it might have changed because the mass-lum relation has changed.)

Thanks for all your careful work on this!

aphearin:
@rmandelb - good thinking about galaxy size. Fortunately, the size (and ellipticity) modeling are formulated in terms of luminosity, so this should not be an issue. I expect that sizes will improve in v4.6, since at fixed luminosity bulges will be ~30% smaller in v4.6 due to the change discussed [here](https://github.com/LSSTDESC/descqa/issues/13#issuecomment-391013338). 

Otherwise, I've been going through my usual checklist for potential downstream consequences of changing <Mr | M*>(z). My local testing suite is reasonably large, and for all the additional validations that I've had time to check by regenerating plots, everything looks good. But it's a large dimensional space with many special subspaces that require TLC, so at some point I really have to lean on the crowd-sourcing of DESCQA (and also on the private RedMapper tests), and leveraging that requires the proper production of a lightcone. 

It's worth briefly mentioning the evolution in <Mr | M*>(z) shown [in this plot above](https://github.com/LSSTDESC/descqa/pull/47#issuecomment-392780748) is almost certainly unrealistically strong. When I add in additional synthetic low-mass galaxies below the UniverseMachine resolution limit, it becomes possible to get away with less aggressive evolution for the low-mass galaxies. I'm mostly unworried about this for cosmoDC2 purposes since <Mr | M*>(z) is not a quantity of intrinsic scientific interest to WGs. I'd like to try and get away without that additional modeling ingredient for v4.6, but now that this flexibility is in the model, I expect it will be less arduous to recalibrate this evolution in future iterations that implement the addition of synthetic low-mass galaxies (which, incidentally, looks very promising for achieving a proper power law scaling all the way to r<28).

evevkovacs:
@duncandc @yymao @aphearin We have a new version of protoDC2 (v4.6) which has been retuned to pass dn/dmag, but we need to check it. This test is not yet in the master branch of descqa. Is it likely to be soon (ie today), or can you point me to a branch  I could clone so I can run the test?  Actually, please point me to branches of any other tests that are close to done, but not yet in master. For example, I see that the CLF test has been updated in master, so is that now final? Thanks very much!

yymao:
@evevkovacs It's really difficult to define what is *final* but the CLF test (with true halos) is the master branch is indeed *current*. 

I think @duncandc is updating this dn/dmag test and you can get the *current* version by running
```sh
git checkout -b duncandc-devel master
git pull https://github.com/duncandc/descqa.git devel
```

duncandc:
I am fiddling with it right now, but it currently works and can be run following @yymao 's directions.

evevkovacs:
@duncandc @yymao @aphearin Thanks very much, the results are [here]( https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-06_10) Although, formally, we fail the tests, visual inspection of the plots leads me to conclude that the agreement is acceptable (and meets the target of falling below the requirement in anticipation of the increase in dn/dmag due to the increase in the redshift range for cosmoDC2 )

duncandc:
Thanks for running that.  It looks like the bottom difference panel is not showing what it should.  I will fix that.

yymao:
@evevkovacs thanks! The catalog seems be to doing pretty good. 

I believe we discussed about loosening criteria; maybe @duncandc is still working on that? Another aesthetic suggestion for @duncandc: making the grey band lighter/more transparent.

duncandc:
The criteria has been loosened, but it is easily changed to whatever we want in the test by changing the tolerance or magnitude range arguments.

evevkovacs:
I think the catalog is doing very well, but wait for @aphearin to have the final word.

yymao:
@duncandc what's the current criteria? 

I think @rmandelb's earlier suggestion was "20%-level match for 23 < i < 25.3, and 40%-level match for 25.3 < i < 27". But given what we've learned, maybe we should be between 0 and -50% for 23 < i < 26.5? 

(cc @aphearin)

aphearin:
This level of agreement is what I anticipated based on the calibration experiments appearing [earlier in this thread](https://github.com/LSSTDESC/descqa/pull/47#issuecomment-393399169). The turnover at r>26 is expected, as we decided to postpone calibrating model components that will rectify this until we have cosmoDC2 cutouts to work with, as discussed with @rmandelb and @janewman-pitt-edu.

evevkovacs:
@duncandc @yymao We are planning to have a final version of protoDC2 v4.6 ready in the next day.
Is there any possibility that the dn/dmag test could be in the master branch by late tomorrow so we could do a full descqa run of all the required validation tests? Otherwise let me know if there are updates to that development branch that I will need to pull. Thanks Eve

duncandc:
sure, I will put in a pull request this afternoon.

duncandc:
ok @yymao this should be ready to go.  You can see an example run here: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-07_6

@aphearin let me know if you want something changed.

evevkovacs:
@duncandc @yymao @dkorytov @danielsf I was curious as to why the y-band was being skipped. Dan looked into it. In the reader, the quantity is mag_true_Y_lsst, not mag_true_y_lsst (see line 260 in alphaq.py) (and similarly for all the other Mag quantities). It looks like this might be an issue in buzzard too, since it is also skipped. Which of these is actually correct? Is it y or Y? Did we spell this out in the schema? Thanks.

evevkovacs:
@duncandc @yymao  If I want to run the latest version of the test, how do I get the code? I tried a git pull on the cloned branch that I made yesterday, but it said I was up to date. Thanks.

yymao:
@evevkovacs 
```
git checkout duncandc-devel
git pull https://github.com/duncandc/descqa.git devel
```

yymao:
@aphearin @rmandelb I wonder if you have further comments on this test? If not, I'd like to merge this.

aphearin:
For posterity's sake, I'm posting an updated plot showing <Mr|M*, z> in cosmoDC2 v1.0.0. This is to be contrasted with the protoDC2 result that I posted [earlier in the comment above](https://github.com/LSSTDESC/descqa/pull/47#issuecomment-392780748). My earlier plot showed that protoDC2 had an unphysically strong redshift evolution at the faint end, revealed to be grossly inconsistent with the COSMOS data shared by @rmandelb that appears in [this comment](https://github.com/LSSTDESC/descqa/pull/47#issuecomment-392293176) of this thread. Modeling <Mr|M*, z> with reasonable accuracy turned out to be mission-critical for simultaneously fitting [HSC dn/dmag](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-09-10_12&test=ApparentMagFuncTest_HSCi&right=2018-09/2018-09-10_12/ApparentMagFuncTest_HSCi/summary.png) and [DEEP2 dn/dz](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-09-09_27&test=Nz_r_DEEP2_JAN&right=2018-09/2018-09-09_27/Nz_r_DEEP2_JAN/summary.png). So thanks again to everyone here for the hard work in nailing down these validation criteria. 

mr_mstar_evolution

PR #51, opened 2017-12-21
Stellar mass function test

evevkovacs:
The last 2 functions in the test could go into a utilities module as they are more generally useful. Here's the link to the last run before the pull request.
[]https://portal.nersc.gov/project/lsst/descqa/v2/?run=2017-12-20_36&test=SMF_PRIMUS

PR #52, opened 2017-12-21
add utils module

yymao:
This PR adds a new ` utils` module into `descqa`. Right now this module includes some functions needed in #38 and #51.


evevkovacs:
Not sure if this is where I am supposed to respond, but I reviewed the code and the functions from the SMF test look ok. I will have to change the calling sequence for get_sky_volume once this is committed.

PR #54, opened 2018-01-04
shear test

patricialarsen:
Shear-Shear correlation function comparison between theory code using pyCAMB and simulation using TreeCorr. This does not yet output a test result or plot and there is some repeated code, but it computes both theory and simulation correlation functions within the DESCQA framework.


yymao:
@patricialarsen thank you for submitting this PR! This is great progress and what needs to be done the next is to migrate the content in your notebook into the DESCQA framework.

Basically you'll need to copy the content in cells `In [11]` and `In [15]` to a python file under the `descqa` directory. And then create a corresponding yaml file under `descqa/configs`. 

Once you are done, you can then try running your test (follow instruction [here](https://github.com/LSSTDESC/descqa/blob/master/CONTRIBUTING.md#how-to-run-descqa)) on NERSC and see your test appear on the web interface! 

Please don't hesitate to contact me if you encounter any issue and/or need more detail instruction.

Thanks again! This is very nice!

descqa/shear_test.py
```diff
@@ -0,0 +1,233 @@
+
+#from __future__ import unicode_literals, absolute_import, division
+import os
+import numpy as np
+from .base import BaseValidationTest, TestResult
+from .plotting import plt
+
+
+import sys
+sys.path.insert(0, '/global/common/software/lsst/common/miniconda/py3-4.2.12/lib/python3.6/site-packages')
```
yymao: remove both lines with `sys.path.insert` (this will be taken care by the framework)

descqa/shear_test.py
```diff
@@ -0,0 +1,233 @@
...
+
+from GCR import GCRQuery
+import GCRCatalogs
+import descqa 
+
+from builtins import str
+import yaml
+import numpy as np
+import healpy as hp
+from descqa.plotting import plt
```
yymao: remove this line as it's already imported above

descqa/shear_test.py
```diff
@@ -0,0 +1,233 @@
...
+
+from descqa import BaseValidationTest, TestResult
```
yymao: remove this line as it's already imported above

descqa/shear_test.py
```diff
@@ -0,0 +1,233 @@
+
+#from __future__ import unicode_literals, absolute_import, division
+import os
+import numpy as np
+from .base import BaseValidationTest, TestResult
```
yymao: reorder all imports --- (1) built-in modules, (2) installed modules, (3) this module (those start with `from .`)

descqa/shear_test.py
```diff
@@ -0,0 +1,233 @@
...
+        
+        data = np.random.rand(10) #do your calculation with catalog_instance
+
+        fig, ax = plt.subplots()
+
+        for ax_this in (ax, self.summary_ax):
+            ax_this.plot(data, label=catalog_name)
+
+        #self.post_process_plot(ax)
+        #fig.savefig(os.path.join(output_dir, 'plot.png'))
```
yymao: uncomment these so it'll save the figure to be viewed on the web interface

yymao:
sorry, I accidentally clicked approve...

descqa/shear_test.py
```diff
@@ -0,0 +1,235 @@
...
+
+import sys
+sys.path.insert(0, '/global/common/software/lsst/common/miniconda/py3-4.2.12/lib/python3.6/site-packages')
+sys.path.insert(0, '/global/common/software/lsst/common/miniconda/py3-4.2.12/bin/python')
+sys.path.insert(0, '..')
+
+
+
+from GCR import GCRQuery
+import GCRCatalogs
```
yymao: remove this line as this will be taken care by the framework

patricialarsen:
Thanks @yymao for the notes - should I be testing this on the forked repository with: 
./run_master.sh -v -c proto-dc2_v2.0_test -t shear 
?

yymao:
@patricialarsen yes. you can run the script in your cloned forked repo. You may want to use `-c protoDC2` for the up-to-date version.

patricialarsen:
@yymao everything is working now, the reason this failed the build is because in the run_master script it uses "/global/common/cori/contrib/lsst/apps/anaconda/py3-envs/DESCQA/bin/python" when the camb and treecorr modules are located in 
"/global/common/software/lsst/common/miniconda/py3-4.2.12/bin/python"

I didn't want to make changes to the run_master script without first confirming with you, but when I swap the python environment it works and uploads the results to the website as expected.

yymao:
@patricialarsen thanks! I'll review the code soon.

The python env in `run_master.sh` in the master branch has been already updated (See
https://github.com/LSSTDESC/descqa/blob/master/run_master.sh#L10)

patricialarsen:
Here's a preview of the test within the framework. It's for one bin around redshift 0.5-0.6 for protoDC2. Just the theory v simulations plots for now (x-axis is separation in arcminutes and y-axis is real-space correlation function), there's still no pass-fail criteria.

https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-05_48&test=shear&catalog=protoDC2

rmandelb:
Hi @patricialarsen - I looked at the test preview and have a few questions:

- Are these the standard xi+/xi- shear correlation functions, or something else?  They are labeled w+ and w- but I think they must be the projected xi+/xi-, e.g. as in equation 32 of http://adsabs.harvard.edu/abs/2015RPPh...78h6901K ?

- What is your vision for how this test will be incorporated eventually?  Right now you've just shows a narrow redshift slice, I gather (maybe for speed when debugging?).  Will you eventually use a few broader bins spanning a range of redshifts?

- I think it might be useful to put vertical lines indicating the min/max scales for which shear-shear measurements are typically carried out (e.g., from KiDS or DES).  We would probably not care too much about discrepancies that are well outside that range of scales.

- The vertical axis/tick labels are too wide so they need some tweaking (esp once you add the axis labels).

patricialarsen:
Hi @rmandelb,

Thanks for your feedback. Yes, these should be xi_plus  and xi_minus. Sorry for the labelling confusion. And thanks for your suggestions on the plot - I'll incorporate those in the next version. 

For the extent of this test, it really depends how much testing we want on the catalogues at this verification stage. I can see it either being an initial validation test to make sure the catalogue is roughly okay (i.e. shears don't have any major bugs), or a test to see if the required levels of accuracy of the shear correlation functions are met.  The current narrow redshift bin is just for testing and as the limits are input parameters in the .yaml file it's straightforward to broaden these or to use multiple bins.  

After I've determined some pass/fail criterion, tidied up the output and fixed the current version conflicts, it would be easy enough to make this incorporate:
- cross-correlations between redshift bins
- a check for shear sign convention mismatch
- consistency with convergence autocorrelations
Longer term we could also update the theory to the WLPipe theory code, once we have the necessary modules in the DESCQA environment. I'm very open to suggestions about what would be useful to incorporate (or what wouldn't be useful) if you have any thoughts on this.

evevkovacs:
@patricialarsen  Thanks very much. I have one small cosmetic comment. It is difficult to distinguish the blue and green colors on the w+/xi+ plots. Perhaps a darker value or a different choice for one of the colors would help.

yymao:
@patricialarsen Just checking in --- what's the current status of this test? Is this ready to be review?

patricialarsen:
Yes, the cosmology ideally should be set to that of the catalog. I can update that assuming there's some way to query each catalog's cosmology? In the meantime since it's currently a rough check I think it's still useful to compare to some realistic cosmology. 

I'll remove commented out code - but will wait until you're done with changes so we don't conflict.

yymao:
I think this PR is ready to be merged. Before merging I'll run the latest version on NERSC again to make sure there's no issue.

patricialarsen:
Example output of current version: 

https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-02-23_3
(note that default has no jack-knifing)

PR #58, opened 2018-01-13
Add readiness test

### LINKED ISSUE #57: readiness test (to catch bugs before catalog goes into image pipeline)
Now that we are preparing for Run 1.1, we should make sure there is no obvious bug (like incorrect units, labels, etc) in the catalog before the catalog goes into the image pipeline. @evevkovacs @katrinheitmann and I think it would be a good idea to have a "readiness test" that checks if the quantities that are used by CatSim have reasonable distributions (for example, max, min, mean, std, and histogram for visual inspection).

This test needs to be done by 1/19 (hopefully well before that).

yymao:
This PR will fix #57. 

This PR is **not** yet ready to be merged. For now it is used to track progress on the readiness test and people can use this PR to suggest features and modifications to this test. Rapid development will take place on this branch.

Thanks to @evevkovacs, a preview of current version can be found at https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-12_20&test=Check_ellipticity


aphearin:
@yymao and @evevkovacs - This is an extremely useful and important set of tests. Is there a prioritized list of quantities for which this will be implemented? The code is sufficiently clear that it would be easy for me to pick from the list and match the pattern shown here

yymao:
@aphearin I think we will prioritize those quantities that are used by CatSim.

evevkovacs:
The code is setup to take an arbitrary list of quantities, or else to use a keyword that triggers checks of a particular set of quantities.  I'm working on some improvements to consolidate the plots. I think it will be possible to make the test work both for "readiness" for image simulations as well as a way for users to check particular quantities of interest.

yymao:
@evevkovacs I am ready to merge this PR. Let me know if you have any objection!

evevkovacs:
@yymao @dkorytov  @danielsf I ran the readiness code and checked the result. The size_bulge* finite fraction is still computed without removing the zeros. There could be a parameter in the yaml file called 'filter_zeros' or similar to deal with this. It is not a critical change but would be nice to have. 
Rv and Av are currently computed by CatSim from the B- and V-band luminosities and the extincted luminosities. Scott must be using native quantities. We need a "gcr" quantity to compute these so they can be checked. I think it is just a couple of lines of code to go from the luminosiites to Rv and Av. Rv and Av are in the schema, so they need to be added to the reader for protoDC2. This would fix the problem of missing Rv and Av.
 The hovering doesn't work for me on linux firefox either, but I don't think it is critical. 
Otherwise, everything looks good.

PR #66, opened 2018-01-24
update readiness test (add SED, acceptable ranges, relation checking)

### LINKED ISSUE #60: finalize the acceptable quantity ranges for the readiness test on protoDC2
Now that we have a readiness test in place (you can [see a demo here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-18_9&test=readiness_protoDC2&right=2018-01-18_9/readiness_protoDC2/protoDC2/SUMMARY.html)), we need to finalize the acceptable ranges for each quantities in protoDC2, especially those used by the image generation. 

All the acceptable ranges are set in [this YAML config file](https://github.com/LSSTDESC/descqa/blob/master/descqa/configs/readiness_protoDC2.yaml). @evevkovacs and I have put in many quantities, but I don't think we have exhaust them. Some of them do not have have acceptable ranges specified. 

Once this is done, it means we can sign off protoDC2 2.1.2. I assign this to @evevkovacs and @dkorytov but I think we might also need help from @danielsf and @abensonca. Also cc'ing @rmandelb @katrinheitmann @jchiang87 for their information.

### LINKED ISSUE #64: Add SEDs to readiness test
@yymao  Please take a look at https://github.com/evevkovacs/descqa/blob/ready_SEDS/SED_test.ipynb
The distributions are not as unreasonable looking as I expected, if we convert the SEDs to magnitudes. I am thinking that we can add a keyword to the yaml file such as function: -2.5*np.log10, which tells the readiness test to first evaluate the function on the catalog data before plotting it or computing any other statistics.  What do you think?

yymao:
This PR updates the readiness test for protoDC2 with the following features:
- including SEDs (fixes #64)
- add more acceptable ranges (fixes #60)
- add a new feature to do simple relation checking

➡️ [Click here to see the test result from this PR](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-24_2&test=readiness_protoDC2&right=2018-01-24_2/readiness_protoDC2/protoDC2/SUMMARY.html) 

Can @evevkovacs @dkorytov @abensonca comment on whether the results look good? Currently the catalog is failing because some of the SED std is too big, but if that's normal we can change the acceptable ranges. 

Also, any other simple relations we should include? I only added two as a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ script:
 - nosetests
 - python -c "import descqaweb"
 - python -c "import descqarun"
-- '[ -z "$CHANGED" ] || pylint --disable=C0103,C0301 --extension-pkg-whitelist=numpy $CHANGED'
+- '[ -z "$CHANGED" ] || pylint --disable=C0103,C0301 --extension-pkg-whitelist=numpy $CHANGED; EXITCODE=$?; if [ $(($EXITCODE&7)) -gt 0 ]; then exit $EXITCODE; fi;'
--- a/descqa/configs/readiness_protoDC2.yaml
+++ b/descqa/configs/readiness_protoDC2.yaml
@@ -137,6 +137,10 @@ quantities_to_check:
     f_zero: 0
     f_outlier: [0, 0.05]
 
+  - quantities: 'stellar_mass*'
+    log: true
+    f_nan: 0
+
   - quantities: 'sed_*_*[0123456789]'
     label: SED
     log: true
@@ -144,21 +148,33 @@ quantities_to_check:
     max: [9, 11]
     mean: [5, 8]
     median: [5, 8]
-    std: [0.5, 0.7]
+    std: [0.5, 1.3]
     f_nan: 0
     f_inf: 0
     f_zero: 0
     f_outlier: [0, 0.03]
 
   - quantities: 'sed_*_*_disk'
-    label: SED
+    label: Disk SED
     log: true
+    f_nan: 0
+    f_inf: 0
+    f_zero: 0
+    f_outlier: [0, 0.03]
 
   - quantities: 'sed_*_*_bulge'
-    label: SED
+    label: Bulge SED 
     log: true
+    f_nan: 0
+    f_inf: 0
+    f_zero: 0
+    f_outlier: [0, 0.03]
 
 
 relations_to_check:
  - 'size_minor_bulge_true <= size_bulge_true'
  - 'size_minor_disk_true <= size_disk_true'
+ - '(size_bulge_true != 0) | (stellar_mass_bulge == 0)'
+ - 'stellar_mass_bulge <= stellar_mass'
+ - 'stellar_mass_disk <= stellar_mass'
+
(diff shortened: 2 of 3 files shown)


abensonca:
For the failing SED tests I notice a couple things:

1) The failures are the 4 shortest wavelength bins;

2) Based on the min/max/median/mean values it looks to me that the large stdev for these is being driven by a low min.

Given that, I don't find this too surprising - the SED at those short wavelengths is dominated by short-lived stars. So, for example, galaxies which get their star formation quenched are going to fade at those wavelengths quickly, leading to low values for the SED there. So I think a large stdev is not unreasonable for these SED bins.

yymao:
Thanks @abensonca, I'll update the acceptable ranges then. 

@evevkovacs @dkorytov (or anyone else) do you have any other comments before I merge this?

evevkovacs:
@yymao Can we order the SEDs in increasing or decreasing wavelength? I emailed you about a color gradation idea for the SED plots. Happy to pursue that. What about Av and Rv? Those should be in the reader now. I can work on that too.

yymao:
@evevkovacs I disabled legends for plots that have more than 9 quantities. The legend is not really helpful in that case. 

SED labels are not easily sortable given the way we name them. We could change our schema to require all wavelengths 6-digit numbers with preceding 0's in the labels..., but this is not what we are doing right now. 

I don't think `A_v` and `R_v` are in the reader yet (haven't seen any PR), but we should move that discussion to https://github.com/LSSTDESC/gcr-catalogs/issues/70

evevkovacs:
I tried the sorting. See https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-24_9&test=readiness_protoDC2&right=2018-01-24_9/readiness_protoDC2/protoDC2/SUMMARY.html
I think it makes the SED information easier to read.

evevkovacs:
@yymao pushed code with sorted quantities

yymao:
@evevkovacs thanks. I just renamed the function and move it outside the class (since it is not specific to this class).

yymao:
➡️ [Updated test result](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-25_2&test=readiness_protoDC2&right=2018-01-25_2/readiness_protoDC2/protoDC2/SUMMARY.html) (corresponding to 6cd6fc1) 

@abensonca can you comment on whether the SEDs for bulge and disk look reasonable? (See last two plots. Don't worry about the inf) 

@abensonca @evevkovacs @dkorytov For the disk/bulge SEDs/stellar masses, do you think it makes sense to set them to zero if they are smaller than a threshold? There are some tiny values and I am not sure if they are physical. (I am not necessarily suggesting doing this for Run 1.1, of course).

evevkovacs:
@yymao @dkorytov @abensonca  For the galsampled version of protoDC2 and will be vetting the galaxies to make sure that we don't sample from obviously anomalous values. But that said, very small disk and bulge stellar masses are possible. I think as long as the total stellar mass and total luminosities are sensible, there isn't a problem, but @abensonca should comment.

abensonca:
@yymao Yes - those look reasonable to me.

@evevkovacs I agree - if the total stellar mass and luminosities look ok we shouldn't be concerned about very small values for the individual components.

yymao:
@evevkovacs I am going to merge this. Let me know if you have any objection.

evevkovacs:
Looks good.

Issue #77, opened 2018-02-16
galaxy size distribution


yymao:
This is a required validation test from the WL WG @msimet. This test is to check the distribution (mean, scatter, etc) of galaxy sizes. This test is somewhat related to the size-luminosity test #13. 

Proposed validation data set is COSMOS. 

- [ ] code to reduce mock data
- [ ] code that works within DESCQA framework
- [ ] validation data
- [ ] validation criteria

PR #80, opened 2018-02-23
Size distribution validation test

msimet:
Here's a draft of the size distribution validation test (#77).  The only thing we really care about in the WLWG (beyond the gross characteristics like "order of magnitude is right") is the small-end slope, so I've coded up a test that computes this. As a comparison data set, I computed the size distribution of the Sersic fits to galaxies in the COSMOS postage stamp data set used for Great3; the test re-computes the slope for the same range of sizes used in the catalog it's examining. This may not work if the size is given in something other than arcsec, though, which is the current assumption.

I wasn't sure if it was desirable to actually code up a pass/fail `TestResult` so right now the tests all automatically pass if they make it to the return value; this can be changed of course. And I'm happy to make any other necessary changes as well.

A run of this test and its outputs can be found at https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-02-23_2&test=size_dist_COSMOS


yymao:
@msimet thank you for the PR! The test looks good. There's a few places in the code that need to be fixed to pass the build, and I'll can help with that.

descqa/SizeDistribution.py
```diff
@@ -0,0 +1,124 @@
...
+
+    #plotting constants
+    validation_color = 'black'
+    validation_marker = 'o'
+    default_markers = ['v', 's', 'd', 'H', '^', 'D', 'h', '<', '>', '.']
+    msize = 4 #marker-size
+    yaxis_xoffset = 0.02
+    yaxis_yoffset = 0.5
+
+    def __init__(self, observation='', **kwargs):
```
yymao: `observation` is not used and should be removed.

descqa/SizeDistribution.py
```diff
@@ -0,0 +1,124 @@
...
+
+        #validation data
+        validation_filepath = os.path.join(self.data_dir, kwargs['data_filename'])
+        self.validation_data = np.loadtxt(validation_filepath)
+        
+        self.acceptable_keys = kwargs['possible_size_fields']
+
+        #check for valid observations
```
yymao: I don't think Lines 32-41 are actually used in this test, do they? If not, remove them.

descqa/SizeDistribution.py
```diff
@@ -0,0 +1,124 @@
...
+    #plotting constants
+    validation_color = 'black'
+    validation_marker = 'o'
+    default_markers = ['v', 's', 'd', 'H', '^', 'D', 'h', '<', '>', '.']
+    msize = 4 #marker-size
+    yaxis_xoffset = 0.02
+    yaxis_yoffset = 0.5
+
+    def __init__(self, observation='', **kwargs):
+
```
yymao: add the following line here to pass pylint check.
```python
    #pylint: disable=W0231
```

descqa/SizeDistribution.py
```diff
@@ -0,0 +1,124 @@
+from GCR import GCRQuery
```
yymao: This is not used and should be removed.

descqa/SizeDistribution.py
```diff
@@ -0,0 +1,124 @@
...
+                [1.0*np.sum(N[-i-1:]) for i in range(len(N))], dtype=float
+            )[::-1]/np.sum(N)
+        data_slope, data_intercept = np.polyfit(size_pts[mask], cumul_N_norm[mask], 1)
+        
+        # Compute the slope for the validation dataset in this size range.
+        # Copy the validation dataset so we can play with it
+        validation_data = self.validation_data.copy()
+        validation_mask = (validation_data[:,0] > min_sizes) & (validation_data[:,0]<median)
+        validation_data[:,1] /= validation_data[validation_mask,1][0]
+        validation_slope, validation_intercept = np.polyfit(validation_data[validation_mask, 0],
```
yymao: `validation_intercept` is not used. You can change it to `_` or add `#pylint: disable=unused-variable` to this line to pass pylint check.

descqa/SizeDistribution.py
```diff
@@ -0,0 +1,124 @@
...
+        cumul_ax.set_ylabel("N({})".format(key))
+
+        with open(os.path.join(output_dir, 'size_distribution_{}.txt'.format(catalog_name)), 'w'
+                 ) as f:
+            f.write("# Slope, intercept\n")
+            f.write("%7f  %9f\n"%(data_slope, data_intercept)) 
+
+        fig.savefig(os.path.join(output_dir, 'size_distribution_{}.png'.format(catalog_name)))
+        plt.close(fig)
+        return TestResult(0, passed=True)
```
yymao: Change this line to 
```python
return TestResult(inspect_only=True)
```
because we don't have validation criteria yet.

msimet:
Okay, I think I have addressed these changes. Thanks for the detailed review! Happy to make any other changes you'd like, as well.

PR #81, opened 2018-02-24
Ellipticity

evevkovacs:
Plots for 2 variants of this test are  here:
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-02-23_24
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-02-23_23
Contributing code and 2 yaml files. First variant plots e distributions for varoiuos morphologies and compares to COSMOS data. Second variant plots default ellipticity distributions for catalog. 
Just tested on buzzard. Throws an error that needs to be fixed, but can be done in parallel with other code improvements.


yymao:
@evevkovacs BTW, are the validation datasets the raw data? I think we should just check in reduced data (like histograms).

evevkovacs:
Hmm.. did I forget  to commit the data? The files came from  BenjaminJoachimi. I don't think they are too big.  Maybe we can just use them for now and decide later?

evevkovacs:
We won't be able to change the binning if we just save the histograms

PR #99, opened 2018-04-07
Readiness tests for instance catalogs

fjaviersanchez:
I modified the readiness tests for protoDC2 to check the instance catalogs given that some columns are not present in these, and most have different column names.


fjaviersanchez:
Here are some results: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-07_12&catalog=instancecatalogs&right=2018-04/2018-04-07_12/readiness_instance/instancecatalogs/traceback.log

katrinheitmann:
This will be really helpful ones all done. It would be good to understand which of the red entries should worry us -- the ra and dec values are of course ok since we rotated the catalog, but what about the rest? Thanks!

yymao:
@fjaviersanchez (cc @katrinheitmann @evevkovacs), I've modified the instance catalog reader so that it now uses the same quantity names as in our schema whenever possible. (You can find the updated reader at https://github.com/LSSTDESC/gcr-catalogs/pull/80). 

So we can run the original readiness test on the instance catalog and here's what I got: 
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-07_26&test=readiness_protoDC2&right=2018-04/2018-04-07_26/readiness_protoDC2/instance_test/SUMMARY.html

I'm going to close this PR without merging it for two reasons: (1) we don't need the new readiness test now that we updated the reader and (2) this PR contains #87 ...

We should move to https://github.com/LSSTDESC/gcr-catalogs/pull/80 to discuss instance reader.

Issue #73, opened 2018-02-14
color-dependent galaxy clustering


yymao:
As discussed in #10 and #63, we'd like to make sure the catalog has reasonable color-dependent galaxy-galaxy clustering signals. 

This should be relatively easy to implement: we can use the current galaxy-galaxy clustering and just implement color-selection. We can use SDSS data as validation data. Validation criteria TBD. 

- [x] code to reduce mock data
- [x] code that works within DESCQA framework
- [x] validation data
- [ ] validation criteria

yymao:
Implemented in #100. Given the small sky area of protoDC2, having validation criteria is not meanfuling at this moment. We can revisit this if we have snapshot data / a larger sky area.

PR #100, opened 2018-04-08
add color-dependent tpcf test

yymao:
This PR adds the color-dependent tpcf test (as requested in #73). A run of this test on the protoDC2 catalog can be found at:
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-07_24&test=tpcf_Zehavi2011_color

While the code works, there seems to be some issues/bugs with the correlation calculation. However, I still wasn't able to identify the issue. So I'm still submitting this PR and hopefully people can take a look. 

(cc @evevkovacs @vvinuv @aphearin)


yymao:
@aphearin @evevkovacs I think I understand why the correlation functions look strange in some cases -- I don't think we have enough galaxies for some of these samples given the redshift cuts we applied! In particular, I now print out the galaxy count in each sample and you can see that [these numbers] are not large enough to produce robust measurement. If we want to test low-z clustering maybe we need to do it on snapshots?

aphearin:
What is the redshift range @yymao - these numbers are awfully low?

yymao:
We are using the redshift ranges as in Zehavi 2011. However, due to the small sky area of protoDC2, there's very few galaxies in those low redshift bins (volume too small). So I think we'll have to test this on snapshot catalogs (unless there's other smart suggestions).

evevkovacs:
@yymao @aphearin  But I recall the test results also didn't look good for Buzzard  and that catalogue should not suffer from lack of galaxies. We should run Buzzard through all of the latest versions of these correlation-function tests to see where we stand.

yymao:
@evevkovacs I don't think we've run tpcf on the full Buzzard --- we might have to submit a job for that.

evevkovacs:
I think it would be worth doing. Run all of the available tpcf tests

PR #102, opened 2018-04-10
redmapper conditional luminosity function test

### LINKED ISSUE #9: testing luminosity function of cluster galaxies (satellites, centrals)
@rmandelb is the idea to do a conditional luminosity function using true halos? 

- [x] code to reduce mock data
- [x] code that works within DESCQA framework
- [ ] validation data
- [ ] validation criteria

chto:
I have implemented a conditional luminosity function measurement on redmapper catalog. [as discussed in #9 ] The run on protoDC2-v2 redmapper catalog looks like
clf_redmapper
[I somehow got permission denied to the DESCQA web directory, so I cannot post the link ]. 

Pining @j-dr @yymao @erykoff as interest parties. 

[Edited by @yymao: This PR fixes #9 ]


yymao:
@chto thanks. I haven't looked at the code in detail but wonder if this test can be consolidated with the existing code in `descqa/clf_test.py`?

chto:
@yymao  I am not sure.  descqa/clf_test.py mainly deals with halo catalog, while this one deals with redmapper catalog, which requires consideration of limiting magnitude, membership probability, etc. 
I think this is a different level of test on simulations, so it's probably better to keep them separate.

yymao:
@chto hmm, ok. I'll have to take a closer look at the code to be sure. My thought was that test can use different sources (true halos or redmapper groups) as halos. In the case of true halos, the membership probabilities are just 1 and 0. 

Anyway, I'll come back to this tomorrow.

yymao:
@chto also, can you clean up the code a bit. In particular,
- removing testing statements like `print('test2')`
- enforcing a consistent coding style (e.g., have spaces before and after operators)

Also, there's a large number of `for i in range(...)` statements. In most cases these can be avoided (either by directly looping over the iterable itself, or by using numpy functions). If it's helpful, I can help rewrite some of these to give you an example.

chto:
@yymao Sure. I will try to make this more readable shortly.

chto:
@yymao I tried to make the code more readable and tried to avoid "for i in range" statement. But, there are still places where I need the index of the array, so I cannot avoid that statement completely.

yymao:
@chto thanks. I'm traveling this week but will try to find time taking a look at your updates.

katrinheitmann:
Just curious: is this comparison really against v2? Can we do it for v3 
instead? Thanks!

chto:
@yymao. I fix a bug in absolute magnitude calculation as well as add kcorrection while converting apparent magnitude to absolute magnitude. Besides, I add a validation criteria, which seems to be able to distinguish protodc2-v4.3 and protodc2-v4.5.
 
The latest result is here https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-05-31_8. 
I think the code is in pretty good shape now. Please do take a look.

descqa/kcorrect_wrapper.py
```diff
@@ -0,0 +1,26 @@
+import kcorrect as kc
+import numpy as np
+from tqdm import tqdm
+def kcorrect(mags, magerrs, redshifts, bandshift, filters="des_filters.dat"):
+   kc.load_templates()
```
yymao: Please always use 4 spaces for indentation.

descqa/kcorrect_wrapper.py
```diff
@@ -0,0 +1,26 @@
...
+   kc.load_filters(f=filters)
+   maggies=10.**(-0.4*mags)
+   maggies_ivar= 1./(maggies*0.4*np.log(10)*magerrs)**2 
+   kcorrect_arr = np.zeros_like(magerrs)
+   pbar = tqdm(total=len(mags))
+   for i, mag in enumerate(mags):
```
yymao: use zip or stack `magerrs`, `maggies`, and `maggies_ivar`. Don't use `i` unless the index value itself it really needed.

descqa/clf_redmapper.py
```diff
@@ -0,0 +1,513 @@
...
+
+    def prepare_galaxy_catalog(self, gc):
+        # list all quantities that is needed
+        quantities_needed = {'cluster_id_member', 'cluster_id', 'ra', 'dec', 'ra_cluster', 'dec_cluster',
+                             'richness', 'redshift_true', 'lim_limmag_dered', 'p_mem', 'scaleval', 'p_cen', 'redshift_cluster'}
+        try:
+            magnitude_fields = []
+            magnitude_err_fields = []
+            for band in self.band_kcorrect:
+               possible_mag_fields = [magfield.format(band) for magfield in self.possible_mag_fields]
```
yymao: use 4 spaces for identation

descqa/clf_redmapper.py
```diff
@@ -0,0 +1,513 @@
...
+                                                             pcen=quant['p_cen'][mask],
+                                                             cluster_lm=quant['richness'][mask],
+                                                             cluster_z=quant['redshift_cluster'][mask])
+        ##
+         
+        scores_shift = []
+        scores_scatter = []
+        for i in range(self.n_z_bins):
+            for j in range(self.nlambd_bins):
+               #calculate the relative brightness of centrals and satellites
```
yymao: use 4 spaces for identation

descqa/clf_redmapper.py
```diff
@@ -0,0 +1,513 @@
...
+                for galtype in galaxytype:
+                    if galtype == "centrals":
+                        name = "cen"
+                    else:
+                        name = "sat"
+                    try:
+                        loaded_data = np.loadtxt(self.data_dir + "/clf/{5}/clf_{4}_z_{0}_{1}_lm_{2}_{3}.dat".format(zrange[0], zrange[1], lambdlow, lambdhigh, name, self.compared_survey))
+                        loaded_data[:, 0] += 5*np.log10(h) #kcorrect distmodule assume h=1, so we need to adjust it to match h in simulation
+                        data[galtype].append(loaded_data)
+                    except:
```
yymao: specify exception type

descqa/clf_redmapper.py
```diff
@@ -0,0 +1,513 @@
+from __future__ import unicode_literals, absolute_import, division, print_function
+import itertools
```
yymao: not used.

chto:
@yymao I make some changes to the code, and I think I fix all the comment you wrote above. Please take a look.

yymao (dismissed):
I've edited the code to fix dependency issues and to make it pass quality checks. While this is probably ready to merge, the change set is significant that I think rerunning this test before merging is wise.

evevkovacs:
@yymao I tried to run this by cloning from chto:master (hope that was correct) and got an error:  
Traceback (most recent call last):
  File "/global/u1/k/kovacs/descqa2-local/descqa_chto/descqa/clf_redmapper.py", line 7, in <module>
    import kmeans_radec
ModuleNotFoundError: No module named 'kmeans_radec'

See https://portal.nersc.gov/cfs/lsst/descqa/v2/?run=2020-12-09_4&test=CLF_redmapper&right=2020-12/2020-12-09_4/CLF_redmapper/traceback.log

evevkovacs:
Oh maybe I should have cloned from descqa and then switched to this issue, but I don't see it as a branch? Sorry not sure how to get the code version you want me to test.

yymao:
@evevkovacs This is on Chun-Hao's fork. You can check this branch out by running:

```bash
git checkout -b chto-master master
git pull https://github.com/chto/descqa.git master
```

evevkovacs:
OK, did that and got the same error. https://portal.nersc.gov/cfs/lsst/descqa/v2/?run=2020-12-09_6&test=CLF_redmapper&right=2020-12/2020-12-09_6/CLF_redmapper/traceback.log

evevkovacs:
See https://github.com/esheldon/kmeans_radec.  we have to get this into the path I guess

chto:
@evevkovacs  I think you might want to do `python setup.py install` after git pull.

evevkovacs:
@chto Do you mean in the descqa directory for your branch? I tried your suggestion  and got a bunch of errors. It's trying to install to  /global/common/software/lsst/common/miniconda/py3.7-4.7.12.1-v2/envs/desc/lib/python3.7/site-packages/ which I don't have write access for.  
@yymao  The issue is that this test needs kmeans_radec in https://github.com/esheldon/kmeans_radec. There is a line in /global/homes/k/kovacs/descqa2-local/descqa_chto/setup.py to find this module and install it, but when I try to run it, that module is not being found. How do I fix this? Thanks

yymao:
Request made: https://github.com/LSSTDESC/desc-help/issues/53

rmjarvis:
FYI, TreeCorr has an implementation of kmeans in RA/Dec coords, which is quite a lot faster than Erin's code and produces moderately better results (in terms of having close to uniform patch areas).  cf. https://rmjarvis.github.io/TreeCorr/_build/html/patches.html#running-k-means

It might be easier to just swap out the code for `make_jack_samples_simple`, rather than add a new dependency.  I believe the equivalent code using TreeCorr is:
```
    def make_jack_samples_simple(self, RA, DEC):
        import treecorr
        cat = treecorr.Catalog(ra=RA.flatten(), dec=DEC.flatten(), ra_units='deg', dec_units='deg')
        labels = cat.getNField().run_kmeans(self.njack)
        uniquelabel = np.unique(labels)
        jacklist = np.empty(self.njack, dtype=np.object)
        for i in range(self.njack):
            jacklist[i] = np.where(labels != uniquelabel[i])[0]
        return jacklist
```

evevkovacs:
@chto Would you be able to change your test to use the treecorr implementation of kmeans in RA/Dec coords. (See above) This sound like it would be pretty quick to do. Could you please take a look and go ahead if feasible and update the PR?
Thanks very much.

chto:
@evevkovacs, 
I have implemented @rmjarvis' suggestion with small modification. 
I have run the code and it runs! 

However, I notice that `/global/common/software/lsst/common/miniconda/current/envs/desc/bin/python` doesn't have kcorrect_python installed, which is needed for the module and can be installed by pip.

heather999:
@rmjarvis @chto Is this the correct [kcorrect_python](https://github.com/nirinA/kcorrect_python)?  My attempt to do a pip install failed, just using the tarball.  I can go back and try a manual install as the README suggests.  Do we have some confidence about the long term support of this package?

erykoff:
Oof.  The only `kcorrect_python` that I can find (same as you linked to) is python 2 only.  So that's not going to work with our software.

evevkovacs:
Oh dear. This doesn't seem like a good solution after all. Should we go back to Erin Sheldon's version  https://github.com/esheldon/kmeans_radec

rmjarvis:
kcorrect_python has nothing to do with the TreeCorr kmeans.  This is an orthogonal issue.

rmjarvis:
The last commit to kcorrect_python was over three years ago.  I'd probably recommend not using that and finding some other package that can perform the same (or equivalent) calculation.  Maybe something in astropy?

erykoff:
There is nothing in astropy that does "kcorrections" (TM).  Depends on what functionality we're using on how hard it is to put together a replacement.

rmjarvis:
Yeah, I found some discussions about making an affiliated package for kcorrect, but it doesn't seem to have gone anywhere.  What I did find was the astrophysics package, which claims to implement the Blanton k correction:
https://pythonhosted.org/Astropysics/coremods/phot.html?highlight=kcorrect#astropysics.phot.kcorrect
Furthermore, I think this project is being considered for incorporation into astropy, so that's probably a long term workable option if this does what you need.

rmjarvis:
Ugh.  Never mind.  That project also seems to be defunct now.  :(

rmjarvis:
I can't find anything that does this as part of a maintained python package.  

OTOH, the license for kcorrect_python is listed as "Public Domain", so you could just copy the relevant parts of that code into this repo, make them work for python 3, and then just use that going forward.  That would be better IMO than depending on an unmaintained package.

evevkovacs:
@chto. How critical is to run this kcorrect_python for the test? Could it be dropped? Alternatively, could it be applied one time to the validation data, so that we don't need to run it each time for the test? I am not sure how the module is used, so I am just throwing out ideas for how to get around the issue.

yymao:
Want to echo @evevkovacs's comment above --- When I was cleaning up the test code in this PR, I already made `kcorrect` an optional dependency. It is used only when the k-corrected validation data set does not exist (see [here](https://github.com/chto/descqa/blob/c4cab0b40324aaa883737bfd20283cd5aa266b2e/descqa/kcorrect_wrapper.py#L3) and [here](https://github.com/chto/descqa/blob/c4cab0b40324aaa883737bfd20283cd5aa266b2e/descqa/clf_redmapper.py#L177) and []()). So the k-correction can be run once (outside DESC environment) and the resulting validation data can be checked in to data directory of this repo.  

If approximated k-correction is acceptable, you can also consider using http://kcor.sai.msu.ru/calc_kcor.py  --- This is not a package but a single python function that contains fitting functions for approximated k-correction. The source is [licensed under BSD](http://kcor.sai.msu.ru/license.txt) so we could include it in our source with proper attribution.

chto:
@yymao I think it is the other way around. The k-correction is done on the simulated catalog in the same way as we do on the comparison data. I think maybe we can remove k-correction on both side but there will be an issue if the filters used in the simulation are different from the comparison data (SDSS). 

Further, I don't think the approximated k-correction mentioned above can handle arbitrary filters. 

Here is my proposal: 
1. We add a comparison data without k-correction applied. 
2. If the user has kcorrect-python installed, we do the comparison as it is. If not, we can compare the non-kcorrected simulation luminosity to non-kcorrected data luminosity, and we print a warning message.

yymao:
@chto --- thanks, that sounds like a good solution to me. If you can help implement it that'd be great.

chto:
@yymao Sure. Working on it.

evevkovacs:
@chto @yymao I took a quick look in the code and noticed that "possible_mag_fields" includes only "mag_{}_lsst". (You are trying to compare to SDSS data, and that is why you are doing k-corrections, right?)  But did you know that cosmoDC2 has SDSS magnitudes already available ('mag_{}_sdss')?  I would suggest prepending 'mag_{}_sdss' to the list of possible_mag_fields and if they are available, also skipping the call to kcorrect.  This would be a minor modification to step 1, above.  There are also fits to k-corrections between various filters in [CheckColors.py](https://github.com/LSSTDESC/descqa/blob/master/descqa/CheckColors.py) derived by Nacho Sevilla.
The ones for lsst filters are given in lines 40-56. You could use these for a transformation between lsst and sdss

chto:
@yymao @evevkovacs  I have implemented suggestions above and the code runs on proto-dc2_v3.0_redmapper. However, I also notice that it cannot be run on cosmoDC2_v1.1.4_redmapper_v0.5.7, because the naming conventions are different and also miss a required field, lim_limmag_dered. 

Also,  there is no SDSS magnitudes in cosmoDC2_v1.1.4_redmapper_v0.5.7.

yymao:
Thanks @chto! And I wonder if @erykoff can answer your [questions](https://github.com/LSSTDESC/descqa/pull/102#issuecomment-744460503).

evevkovacs:
@chto So you could add a variable like possible_redmapper_quantities and list the allowed values that work for both redmapper catalogs, and then use the grc function "first_available" to get whichever one is given in the version of the catalog is being used. Please ask @erykoff about lim_limmag_dered; I would think that there is  a substitute or work-around for this quantity.

chto:
@evevkovacs  
Thanks for the reply. I guess I am not entirely sure what is the newest catalog which you want to test. (or alternatively, what are the catalogs that you want to test?)
I was assuming it is cosmoDC2_v1.1.4_redmapper_v0.5.7, but this catalog doesn't have the mag_{}_sdss you mentioned above.

evevkovacs:
@chto Sorry I wasn't clear. I wanted to run on cosmoDC2_v1.1.4_redmapper_v0.5.7 and was suggesting that you use the lsst to sdss conversion in [CheckColors.py](https://github.com/LSSTDESC/descqa/blob/master/descqa/CheckColors.py) to get around the problem of not having mag_{}_sdss available.

evevkovacs (approved):

PR #104, opened 2018-04-10
Img bkg tests

fjaviersanchez:
This pull request tries to incorporate tests on images. Documentation about the tests is still missing and some changes in the validation data are still needed but, hopefully, I'll have this working by tomorrow.


fjaviersanchez:
I managed to make the two tests included work. See here: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-11_1&catalog=focal_plane_0_test

and here:

https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-11_14&test=quick_bkg_pk

Right now, the configurations just pull images from one directory (same as in the case of the instance catalog). Ideally, we would like that the program goes over all the visits available and repeats these tests. A different approach may be just create 10-15 config files with different visits, randomly selected. I think this might be sufficient.

katrinheitmann:
Hi Javier,

The one thing that is very confusing in your set-up is that you call 
this quick-bkg.

Quick background is a setting that is used in PhoSim and has a specific 
meaning and we are not using it currently. I think this could really 
confuse people. Is there a better word/description for this?

Thanks!

katrinheitmann:
Should I worry that the amplitude in Test 33 and R34 is not the same? 
Sorry, I think I am not fully sure what I am looking at ... Thanks again!

fjaviersanchez:
@katrinheitmann Thanks for the comments. I think that you shouldn't be worried because I think that the raft 34 is on the edge of the focal plane so you expect some vignetting and that's the power that we are seeing. The raft that should be compatible with test 30 is R22. This can be changed in the configuration file.

jchiang87:
Raft 11 is also one of the central 9 rafts.

fjaviersanchez:
@katrinheitmann I am suspecting of a bug in the power-spectrum code because I checked with a central raft and the power is still really high... https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-12_1&test=pk_img

fjaviersanchez:
Fixed: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-12_10&test=pk_img (the blue line is perfectly below the orange because I am using the test 9 file to check that the code is working).

And this is checking a different exposure: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-04-12_12&test=pk_img

yymao:
@fjaviersanchez I realized there are too many different things in this PR... can you do the following?

- open a different PR for `descqa/configs/readiness_instance.yaml` --- this one looks fine and I can get it merged quickly.
- remove `descqa/BiasVersusRedshift.py`, `descqa/configs/bias_test.yaml`, and `descqa/configs/readiness_instance.yaml` from this PR

Thanks!

descqa/ImgPkTest.py
```diff
@@ -0,0 +1,83 @@
...
+        if len(test_raft.sensors) != 9:
+            return TestResult(skipped=True, summary='Raft is not complete')
+        xdim, ydim = list(test_raft.sensors.values())[0].get_data().shape
+        total_data = np.zeros((xdim * 3, ydim * 3))
+        # Assemble the 3 x 3 raft's image: Need to use LSST's software to
+        # handle the gaps properly
+        for i in range(0, 3):
+            for j in range(0, 3):
+                total_data[xdim * i:xdim * (i + 1), ydim * j:ydim * (j + 1)] = test_raft.sensors['S%d%d' %
+                                                                                                 (i, j)].get_data()
```
yymao: Lines 35-42: simplify assembling procedure 
```python
        # Assemble the 3 x 3 raft's image
        # TODO: Need to use LSST's software to handle the gaps properly
        total_data = [test_raft.sensors['S%d%d'%(i, j)].get_data() for i in range (3) for j in range(3)]
        total_data = np.moveaxis(np.array(total_data), [0, 1], [2, 3])
```

descqa/ImgPkTest.py
```diff
@@ -0,0 +1,83 @@
...
+        # FFT of the density contrast
+        F1 = fftpack.fft2((total_data / np.mean(total_data) - 1))
+        F2 = fftpack.fftshift(F1)
+        psd2D = np.abs(F2)**2  # 2D power
+        pix_scale = 0.2 / 60 * rebinning  #pixel scale in arcmin
+        kx = 1. / pix_scale * np.arange(-F2.shape[0] / 2, F2.shape[0] / 2) * 1. / F2.shape[0]
+        ky = 1. / pix_scale * np.arange(-F2.shape[1] / 2, F2.shape[1] / 2) * 1. / F2.shape[1]
+        kxx, kyy = np.meshgrid(kx, ky)
+        rad = np.sqrt(kxx**2 + kyy**2)
+        bins = 1. / pix_scale * np.arange(0, F2.shape[0] / 2) * 1. / F2.shape[0]
```
yymao: Lines 49, 50, and 53: (1) use `//` for integer division; (2) use `dtype` to get float array, rather than `* 1.`
```python
np.arange(0, F2.shape[0] // 2, dtype=np.float)
```

descqa/ImgPkTest.py
```diff
@@ -0,0 +1,83 @@
...
+        ax[0].set_ylabel('Number of pixels')
+        ax[0].legend(loc='best')
+        ax[1].plot(bins, ps1d, label=self.raft)
+        ax[1].set_xlabel('k [arcmin$^{-1}$]')
+        ax[1].set_ylabel('P(k)')
+        ax[1].set_xscale('log')
+        ax[1].set_yscale('log')
+        ax[1].set_ylim(1, 1000)
+        self.post_process_plot(ax[1])
+        fig.savefig(os.path.join(output_dir, 'plot.png'))
```
yymao: Add after line 75: close figure
```python
        plt.close(fig)
```

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Returns:
+    --------
+    mean_bkg : Mean background level
+    median_bkg : Median background level
+    bkg_noise: Background noise level
+    """
+    image = image.flatten()
+    q95 = np.percentile(image, 95)  # This is kind of arbitrary but it works fine
+    q5 = np.percentile(image, 5)  # Same as above -> can be substituted by 10 and 90
```
yymao: Lines 27-28: combine them. Also, since the comment suggests these numbers can be changed, do not code the number into variable name
```python
q_low, q_high = np.percentile(image, [5, 95])
```

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+    mask = (image > q5) & (image < q95)
+    median_bkg = np.median(image[mask])
+    mean_bkg = np.mean(image[mask])
+    bkg_noise = np.std(image[mask])
+    return mean_bkg, median_bkg, bkg_noise
```
yymao: Line 29-33: `image` can be directly sliced. Also, simplify the code:
```python
    image = image[(image > q_low) & (image < q_high)]
    return np.mean(image), np.median(image), np.std(image)
```

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def get_predicted_bkg(visit, validation_dataset, db_file, band):
+    if validation_dataset.lower() == 'opsim':
+        return get_opsim_bkg(visit, db_file, band)
+    # TODO add imSim option
```
yymao: raise a `NotImplementedError` for other cases:
```python
    raise NotImplementedError('only "opsim" is currently supported')
```

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def get_airmass_raw_seeing(visit, db_file):
+    conn = sqlite3.connect(db_file)
+    cur = conn.cursor()
+    cur.execute(
+        "SELECT airmass, filtSkyBrightness, finSeeing, rawSeeing, visitExpTime, fiveSigmaDepth FROM ObsHistory WHERE obsHistID==%d"
+        % (visit))
+    rows = cur.fetchall()
+    return rows[0][0], rows[0][1], rows[0][2], rows[0][3], rows[0][4], rows[0][5]
```
yymao: any reason not just do:
```python
    return rows[0]
```
OR
```python
    return rows[0][:6]
```

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def get_opsim_bkg(visit, db_file, band):
+    skybrightness = get_airmass_raw_seeing(int(visit), db_file)[1]
+    mean_bkg = compute_sky_counts(skybrightness, band, 1)
+    median_bkg = mean_bkg
```
yymao: this is a bit confusing; add a comment to explain?

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+
+        fig, ax = plt.subplots(2, 1)
+        ax[0].hist(list(mean_bkg.values()), histtype='step', label='Mean')
+        ax[0].hist(list(median_bkg.values()), histtype='step', label='Median')
+        ax[0].set_xlabel('{} [ADU]'.format(self.label))
+        ax[0].set_ylabel('Number of sensors')
+        ax[1].hist(list(bkg_noise.values()), histtype='step')
+        ax[1].set_xlabel('{} noise [ADU]'.format(self.label))
+        ax[1].set_ylabel('Number of sensors')
+        score = np.mean(np.array(list(median_bkg.values()))) / self.validation_data[0] - 1.
```
yymao: Do we need to convert it to numpy array? It seems that we can use python built-in method here:
```python
   score = sum(median_bkg.values()) / len(median_bkg) / self.validation_data[0] - 1
```

descqa/ImgPkTest.py
```diff
@@ -0,0 +1,83 @@
...
+from .base import BaseValidationTest, TestResult
+from .plotting import plt
+
+__all__ = ['ImgPkTest']
+
+
+class ImgPkTest(BaseValidationTest):
+    """
+    Validation test that computes the power spectrum
+    of a given raft image
```
yymao: Add to docstring the input arguments for this class

descqa/QuickBkgTest.py
```diff
@@ -0,0 +1,144 @@
...
+    mean_bkg = compute_sky_counts(skybrightness, band, 1)
+    median_bkg = mean_bkg
+    bkg_noise = np.sqrt(mean_bkg)
+    return mean_bkg, median_bkg, bkg_noise
+
+
+class QuickBkgTest(BaseValidationTest):
+    """
+    Check of mean, median and standard deviation of the image background.
+    We compare to expeted values by OpSim or imSim.
```
yymao: Add to docstring the input arguments for this class

fjaviersanchez:
Thanks @yymao. I implemented the changes that you suggested. By the way, I would suggest not to run this test on the full focal plane without rebinning (it takes a very long time to run).

yymao:
@fjaviersanchez I've proposed some more changes by submitting a PR to your PR: https://github.com/fjaviersanchez/descqa/pull/1

yymao:
@fjaviersanchez many thanks for bearing with me. I can merge this. The remaining question is whether we need to add some sort of validation data for this test?

fjaviersanchez:
Thank you @yymao. I would say so. Let me please check for a good sample or model.

fjaviersanchez:
@yymao I think that this is PhoSim's cleanest run to get a good estimation of the power spectra: `/global/projecta/projectdirs/lsst/production/DC2/DC2-phoSim-test_WFD-r/output/000035`. What would be the most effective strategy to generate the validation data? Run the script that we have now, dump the output somewhere and re-use it or read these images everytime?

yymao:
@fjaviersanchez that could work. What's the precision that you'd like the PSD to be validated? We could also fit a functional form and use the fitted PSD to validate?

descqa/ImgPkTest.py
```diff
@@ -0,0 +1,132 @@
...
+                msg = 'Raft {} is not complete!'.format(raft_name)
+                ax[1].text(0.05, 0.5, msg, transform=ax[1].transAxes)
+                print('[WARNING]', msg)
+            fig.tight_layout()
+            fig.savefig(os.path.join(output_dir, 'plot_{}.png'.format(raft_name)))
+            plt.close(fig)
+        if count:
+            total_chi2 /= count
+        ndof = len(self.validation_data['k']) - 1
+        score = 1 - chi2.cdf(total_chi2, ndof)
```
yymao: Probably no `1 -` here, since the passing criterion is `passed=(score < 0.95)`? 
```python
score = chi2.cdf(total_chi2, ndof)
```

fjaviersanchez:
@yymao the new validation data should be more representative of most of the images that we'll analyze since I am including sources (the previous one had no sources in the image, that's why it had very low power). I also corrected the bug with the score. Thanks!

yymao:
Thanks @fjaviersanchez! This looks great. Here's latest [DESCQA run](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-08-29_8). I'm merging this now.

PR #113, opened 2018-05-30
Ellipticity validation

msimet:
This code contains the validation criteria for the ellipticity distribution test.  A run of this test on the v4.* catalogs can be found [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-05-29_7).  Everything passes except for the late-type galaxies in proto-dc2_v4.5_rescale, where the 10th percentile is a little high (0.41 instead of <=0.4).  I'm not sure that's actually important--I've left if for now so more knowledgeable people can advise me if that's expected or not. It would probably be okay to adjust that criterion if we needed to, however.

The actual validation criteria used here are:
- min(e)>=0
- max(e)<=1
- 10th percentile < 0.4
- 0.3 < 50th percentile < 0.7
- 90th percentile > 0.6

This is tested for all morphologies separately and for the catalog as a whole; the returned float value is the total number of failed criteria.  The criteria are currently hard-coded but the percentiles could be moved to config files if desired.

### DIFF
--- a/descqa/EllipticityDistribution.py
+++ b/descqa/EllipticityDistribution.py
@@ -361,7 +361,7 @@ class EllipticityDistribution(BaseValidationTest):
         #check overall ellipticity distribution
         global_N = np.sum(np.sum(N_array, axis=1), axis=0)
         number_passed, percentiles = self.validate_percentiles(global_N)
-        print("Percentiles for global distribution are:".format(morphology)+', '.join([" {:.3f} ({})".format(p, v) for p,v in zip(percentiles, self.validation_percentiles['percentiles'])]))
+        print("Percentiles for global distribution are: "+', '.join([" {:.3f} ({})".format(p, v) for p,v in zip(percentiles, self.validation_percentiles['percentiles'])]))
         if number_passed>0:
             print("Ellipticity percentile check failed for global distribution")
         n_fails += number_passed
--- a/descqa/EllipticityDistribution.py
+++ b/descqa/EllipticityDistribution.py
@@ -52,8 +52,8 @@ class EllipticityDistribution(BaseValidationTest):
     }
     
     validation_percentiles = {
-        'percentiles': [10, 50, 90, 100],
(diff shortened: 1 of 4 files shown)


aphearin:
Thanks for the update @msimet - 

CC @dkorytov @evevkovacs - based on the updated DESCQA test Melanie posted above, it looks like the ellipticity model upgrade was applied to v4.4 and v4.5, but not v4.5_rescale.

msimet:
I'd be happy to make that change if you think it's useful!

yymao:
@msimet Yes, in particular because the config files will included in the web interface, it will make it easier for users to find out what validation criteria are used. 

Thank you for making that change!

evevkovacs:
@msimet @aphearin @yymao Thank you Melanie. This looks great. FYI the new ellipticity model has not been applied to proto-dc2_v4.5_rescale.  Another useful feature that could be added a  print out the individual percentile values in the txt file.

msimet:
I think I have implemented these changes--[you can view the new outputs here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-01_22).

descqa/EllipticityDistribution.py
```diff
@@ -326,11 +344,29 @@ def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
...
 
             else:
                 #make empty subplots invisible
                 ax_this.set_visible(False)
                 summary_ax_this.set_visible(False)
+        
+        #check overall ellipticity distribution
+        global_N = np.sum(np.sum(N_array, axis=1), axis=0)
+        number_passed, percentiles = self.validate_percentiles(global_N)
+        print("Percentiles for global distribution are:".format(morphology)+', '.join([" {:.3f} ({})".format(p, v) for p,v in zip(percentiles, self.validation_percentiles['percentiles'])]))
```
yymao: the first `.format` is not actually being used.

msimet: Okay, I think this is fixed now...

PR #118, opened 2018-06-09
SDSS  LOWZ Galaxy-shear test 

EiffL:
This test computes \Delta \Sigma around a sample of lenses selected according to the SDSS LOWZ sample and additional redshift cuts used in Singh et al. (2015) (https://arxiv.org/abs/1411.1755). This is a non trivial test of the galaxy-shear correlation in the simulations, and the actual SDSS measurement is  used for validation (provided by @sukhdeep1989).

This PR is functional, but not complete, additional features include:
  - [ ] validation criteria
  - [x] displaying number density of lenses compared to LOWZ

Here is an example of output on protoDC2 (version 3.0): https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-08_8&test=delta_sigma&catalog=protoDC2

image


EiffL:
@rmandelb Here are plots comparing v3.0 and v4.6.1, I'm a little bit puzzled by why would the lensing signal change between these versions (especially at low separation)
image
image

aphearin:
Thanks for the hard work in implementing this @EiffL. This test is great because it ties so many things together. Could you comment on how the lens sample is selected? I would like to compute DeltaSigma directly from the DM particles in the simulation, which will allow us to decouple the shear calculations from the galaxy-halo connection. If you could provide some details about the lens selection function, that would be helpful. Also, what are the number densities being compared at the title of each figure?

yymao:
@aphearin If I am not mistaken, lenses are `0.16 < redshift < 0.36`, and sources are `i_mag < 26` and `redshift > 0.4` (cf. https://github.com/LSSTDESC/descqa/blob/ca3025c/descqa/DeltaSigmaTest.py#L56)

aphearin:
Yes, pointing me to the relevant lines of code is perfect. Thanks!

rmandelb:
To answer @aphearin 's other question (Francois is traveling at the moment so may be slow to respond):

> Also, what are the number densities being compared at the title of each figure?

The number density is the LOWZ number density.  The first number, which is 14.2 or 10.6/deg^2 depending on the protoDC2 version, is the LOWZ number density in protoDC2 found after the imposition of the selection criteria that Yao pointed you to.  The second number, which is the same in both versions, is the true SDSS LOWZ number density based on the BOSS targeting paper https://arxiv.org/pdf/1509.06529.pdf (table 2).

rmandelb:
I'm going to review the code momentarily.  For those who are not as familiar with this test, let me clarify a few points of interpretation:

As @aphearin says, it ties together several things: 1- that color/mag cuts can give us a realistic sample of bright lenses that is comparable to the ones in real data; 2- that those galaxies have the right galaxy-halo connection (it's possible to get a reasonable large-scale bias without getting the galaxy-halo connection right, so this is an important additional test especially given how DC2 sims will be used for science); and 3- that there isn't some strange bug in the shear calculations.  That is a lot of information all at once.  Andrew's suggestion that he might directly calculate the galaxy-matter cross-correlation would essentially bypass (3).

@aphearin - we could probably get a CMASS test in here as well.  The validation data I have in mind is either https://arxiv.org/pdf/1311.1480.pdf figure 8 (but it has some stellar mass cuts in addition to the basic CMASS cuts, so that's even more complicated) or https://arxiv.org/pdf/1803.08915.pdf figure B1 (orange points in right panel -- simpler sample selection but S/N is lower).  The question is whether this would be worthwhile?  In my view, if there is any concern about LOWZ because of the lower redshift range / smaller volume / some potential issues at low redshift in the light cone, then adding CMASS might be useful.  But if you think it should all work for LOWZ, then it may not be worth the additional work of adding CMASS.

descqa/DeltaSigmaTest.py
```diff
@@ -0,0 +1,118 @@
...
+from astropy.cosmology import WMAP7 # pylint: disable=no-name-in-module
+from .base import BaseValidationTest, TestResult
+from .plotting import plt
+
+__all__ = ['DeltaSigmaTest']
+
+class DeltaSigmaTest(BaseValidationTest):
+    """
+    This validation test looks at galaxy-shear correlations by comparing Delta
+    Sigma to the Singh et al (2015) measurements on the SDSS LOWZ sample.
```
rmandelb: I recommend giving the link directly (http://adsabs.harvard.edu/abs/2015MNRAS.450.2195S or  https://arxiv.org/abs/1411.1755) so people don't have to search.

descqa/DeltaSigmaTest.py
```diff
@@ -0,0 +1,118 @@
...
+        self.comoving_transverse_distance = interp1d(z, d2, kind='quadratic')
+
+    def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
+        # pylint: disable=no-member
+
+        res = catalog_instance.get_quantities(['redshift_true', 'ra', 'dec', 'shear_1', 'shear_2',
+                                               'convergence', 'mag_true_i_sdss', 'mag_true_z_sdss',
+                                               'mag_true_g_sdss', 'mag_true_r_sdss'])
+
+        # Compute mask for lowz sample
```
rmandelb: For future code maintainability, I recommend noting in the comment that this comes from section 3 of the BOSS target selection paper (https://arxiv.org/pdf/1509.06529.pdf) - people may not be familiar with what `lowz' means.

descqa/DeltaSigmaTest.py
```diff
@@ -0,0 +1,118 @@
...
+        # See http://docs.astropy.org/en/v0.3/_modules/astropy/cosmology/core.html#FLRW.angular_diameter_distance_z1z2
+        dm1 = self.comoving_transverse_distance(zl)
+        dm2 = self.comoving_transverse_distance(zs)
+        angular_diameter_distance_z1z2 = u.Quantity((dm2 - dm1)/(1. + zs), u.Mpc)
+
+        sigcrit = cst.c**2 / (4.*np.pi*cst.G) * self.angular_diameter_distance(zs) / \
+                ((1. + zl)**2. * angular_diameter_distance_z1z2 * self.angular_diameter_distance(zl))
+        # Apply unit conversion to obtain sigma crit in h Msol /pc^2
+        cms = u.Msun / u.pc**2
+        sigcrit = sigcrit*(u.kg/(u.Mpc* u.m)).to(cms) / 0.7
```
rmandelb: Hard-coded 0.7's here and below are h's?  I recommend defining `h` and then using it in these equations, for future readability.

descqa/DeltaSigmaTest.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Computing the projected separation for each pairs, in Mpc/h
```
rmandelb: Perhaps clarify here that the validation data is in comoving coordinates (for `r_p` and `DeltaSigma`).

descqa/DeltaSigmaTest.py
```diff
@@ -0,0 +1,118 @@
...
+        gammat = -(res['shear_1'][mask_source][idx2] * np.cos(2*thetac) - res['shear_2'][mask_source][idx2] * np.sin(2*thetac))
+
+        # Binning the tangential shear
+        bins = np.logspace(np.log10(0.05), 1, 17, endpoint=True)
+        counts = np.histogram(r, bins=bins)[0]
+        gt, b = np.histogram(r, bins=bins, weights=gammat*sigcrit)
+        rp = 0.5*(b[1:]+b[:-1])
+
+        counts[counts < 1] = 1
+        gt = gt / counts
```
rmandelb: I recommend issuing a warning if `counts` is below a certain value (like 10 or 100).  It indicates that the test is being misused.

descqa/DeltaSigmaTest.py
yymao: I think `h` should be read in from the catalog's cosmology.

patricialarsen:
@EiffL There are subtle changes between v3 and v4 that could contribute to small scale changes in the galaxy/mass relation. Basically updates are being made to the light-cone code which mostly affect the small scales, and sometimes they're not applied consistently across weak lensing and galaxy positions for various reasons.  I suspect that's what you're seeing here but we're looking into it to make sure. Thanks for flagging this up.

descqa/DeltaSigmaTest.py
```diff
@@ -0,0 +1,118 @@
...
+        # Compute mask for lowz sample
+        cperp = (res['mag_true_r_sdss'] - res['mag_true_i_sdss']) - (res['mag_true_g_sdss'] - res['mag_true_r_sdss'])/4.0 - 0.18
+        cpar = 0.7*(res['mag_true_g_sdss'] - res['mag_true_r_sdss']) + 1.2*((res['mag_true_r_sdss'] - res['mag_true_i_sdss'])-0.18)
+
+        mask_lowz = np.abs(cperp) < 0.2 # color boundaries
+        mask_lowz &= res['mag_true_r_sdss'] < (13.5 + cpar/0.3) # sliding magnitude cut
+        mask_lowz &= (res['mag_true_r_sdss'] > 16) &(res['mag_true_r_sdss'] < 19.6)
+
+        # Counting the number density of LOWZ galaxies
+        nlens = len(np.where(mask_lowz)[0]) / catalog_instance.sky_area
```
aphearin: @EiffL @rmandelb - should this `nlens` count be done here, or after the subsequent 0.16 < z < 0.36 redshift cut on line 57 below?

rmandelb: The code is right - it's before.  The numbers in the target-selection paper do not involve a redshift cut (since it's the target density before spectra are taken).  So Francois had to calculate the LOWZ density before the redshift cut.  But since Sukhdeep's lensing profiles involved a redshift cut, Francois then made the redshift cut before comparing lensing profiles.

aphearin: Ok, just checking, thanks for clarifying

rmandelb: It's good you asked - I think that this should be documented with a comment so that when someone else is trying to update this test a few years from now when running on some other catalogs, they know why it's like this!  Seemingly arcane choices should be documented for the sake of future code writers/maintainers  :)

aphearin: Even better, it would probably be useful to peel apart this into a standalone CMASS selection function, since that is also being used elsewhere to check number densities. I now have my own implementation for model recalibration purposes, so this is no rush so far as I'm concerned, and it's also fine if we stick with the principle of one, two, infinity for now. But the docstring to such a function would be the natural place to document something that's getting reused.

descqa/DeltaSigmaTest.py
EiffL: good point

descqa/DeltaSigmaTest.py
EiffL: added

descqa/DeltaSigmaTest.py
EiffL: Sure, a hard coded `h` is pretty ugly, shame on me... but @yymao reading h from the catalog cosmology would probably mean the whole cosmology should be consistent with the catalog's, no? In the sense that I shouldn't use the default WMAP7 (even if I think that's the right one for DC2)

descqa/DeltaSigmaTest.py
EiffL: yes, this is the sort of things  one can easily forget otherwise (like I did the first time around)

descqa/DeltaSigmaTest.py
EiffL: not sure about a warning, but we can fail the test if the number density gets outside of this  range.

descqa/DeltaSigmaTest.py
yymao: Well, I am not sure what's the best way here, because the catalogs usually won't be in the same cosmology as used in the analysis of the validation dataset. So I guess the answer depends on exactly what you want to compare. If you can compare after you scale little h out, then I guess you'll use catalogs' `h` to scale catalog data and use whatever Singh2015 uses to scale validation data.

descqa/DeltaSigmaTest.py
EiffL: I've added code that will fail the test if the minimum number of bakground galaxies in a radial bin falls below a given number

EiffL:
Thanks everyone for the review and useful comments !

descqa/DeltaSigmaTest.py
EiffL: @yymao I've changed the code so that it will try to read and use the catalog cosmology, and only if that's not defined will it default to WMAP7. sounds good ?

descqa/DeltaSigmaTest.py
yymao: Sure. I am not the expert here who knows how cosmology would impact this validation test, so it's your call.

katrinheitmann:
The cosmology is "WMAP-7 like" -- h=0.71, not 0.7...

patricialarsen:
@EiffL can I suggest printing out the "counts" variable to give a little more information for interpreting the results? I'm wondering how many points are used in the high r_p bins for protoDC2.

yymao:
@EiffL is there any updated that you are still working on? Or is this ready to be merged? 

@rmandelb can you re-review this PR? Thanks!

EiffL:
@yymao  It's all good, I can just add the print that @patricialarsen was asking (in the next couple of hours)

aphearin:
At large r_p, the pair counts should be large. In this regime, noise is from sample variance of the modest box size, not shot noise. So while printing the pair counts is definitely a good thing to do for a sanity check on the validation test, this will not tell us one way or the other about the accuracy of the validation in the larger rp bins.

EiffL:
@patricialarsen I've added a file output to save the number of galaxies per bins.
@yymao , it should be ok to merge, the only thing that is not 100% perfect is that it is missing a validation criteria...

sukhdeep1989:
@EiffL Regarding validation criteria: In CCL, for correlation functions we required that the difference between CCL and benmark should be small relative to the size of expected LSST errorbars. How about something similar here? Or maybe something like a reduced chi^2. I suspect the ProtoDC covariance should be larger than SDSS. I can give you the covariance if you give me the numbers: Volume (~1/20 of SDSS?), source number density (~20X of SDSS?) and lens number density (~1/4 of Lowz?).  With these numbers, simple back of the envelope calculation suggests ProtoDC errors should be about 2X SDSS errors.

yymao:
@sukhdeep1989 your proposal sounds good except that it is potentially too stringent for our purpose. We should allow model error as well as we just want the mock to be reasonably realistic. So maybe we can set 3x or even 4x of SDSS errors?

rmandelb:
I agree with Yao; I'm doubtful that statistical uncertainty in the difference between sims and data should be the driver here.  If we are mostly asking that the test should tell us there is no major *systematic* issue with how galaxies are populating halos, then if the statistical uncertainties of the mock and data added in quadrature are smaller than our tolerance for systematic offsets, the test might end up being too stringent for our needs.  (Also once we move to cosmoDC2, cosmic variance will be far smaller, so the test will suddenly become more stringent if we base it on statistical errors.)

sukhdeep1989:
@yymao Yes, sure. My statement was not about the validation criteria. I was saying that the noise/covariance in ProtoDC measurements should roughly be 2X the noise in SDSS. And that seems reasonable by eye from the plots in this thread. In terms of validation criteria, I think that should be based on how accurate we want the mocks to be. For protoDC, 3X-4X seems reasonable. I presume the ultimate goal will be to have the mocks to within 1X of LSST errors.

descqa/DeltaSigmaTest.py
EiffL: thanks @yymao

EiffL:
Added a fix to prevent seg fault during the call to `search_around_sky`. An option in the configuration will randomly sub sample the background galaxies to a fix number. Without it, I suspect `search_around_sky`was using too much memory, and was crashing on v4.5. 

@yymao It should be ok to merge now, at least it won't segfault anymore :-)

yymao:
@EiffL, thanks! I'm ok with the code (though would appreciate if you use the GCR's native `filters` option to implement the cuts). I do want to make sure someone has reviewed the results though. Can you post a DESCQA run with this updated version? Also, do you or @sukhdeep1989 plan to implement validation criteria as we discussed above?

EiffL:
Here is one run:
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-07-02_4&test=delta_sigma&catalog=proto-dc2_v4.6.1
Apparently, there is no lensing v4.5 and v4.14 but Andrew said that was expected, at least for 4.5, and I don't think v4.14 is official yet.

yymao:
@patricialarsen I wonder if you have comments about this test result that @EiffL posted above?

aphearin:
@yymao - I'm pretty sure that @patricialarsen has been working hard to extend the shear calculations to z>1 for cosmoDC2, and prioritizing this over back-porting things to protoDC2. We will certainly want to put shears in for the "final" version of protoDC2, but at present it seems like cosmoDC2 is the priority. She should comment further on this if I have misunderstood the situation.

EiffL:
Ok sounds good, that's what I was thinking, well, the test is there for whenever a new catalog is ready for testing. BTW, proto-dc2_v4.14 has 51 gal/deg^2 (compared to the 57 deg^2 in lowz), that's pretty good :-)

aphearin:
@EiffL - A big thanks for keeping an eye on this test. Even while the gg-lensing has been non-operational in protoDC2, I have been paying very close attention to the LOWZ number density using your test. So, yes, as version numbers increase, the mock really does improve ;-). Of course the halo mass could still be off, which will show up in the lensing when it is restored. Although it is probably better to further calibrate this halo mass dependence using cosmoDC2 since we will have far better statistics even with the smaller cutouts.

PR #119, opened 2018-06-09
Fix #117 N(z) test errorbar/normalization issue

### LINKED ISSUE #117: bug in dN/dz test: normalizations, errorbars
This issue is for some bugs (at least two) in the dN/dz test.  @yymao and I have been discussing this on slack, but it needs a bit more digging, so we agreed to open an issue (currently assigned to the two of us).

At the DC2 telecon today it was clear that the errorbars on the dN/dz test were wrong (an order of magnitude larger than the scatter in the data points).  Yao and I identified a few things going on:

- This is not a bug, but simply misuse of the test: the latest runs were trying to set the number of z bins using the wrong parameter, so it wasn't getting set (it was setting `Nbins` when it should have been `N_zbins`).  The configuration being used was such that the errorbar determination could not be particularly stable -- it should be the case that number of jackknife regions > number of data points.  This was not the case with the settings that were being used.  He and I have since been exploring use of 15 regions and 8 redshift bins, which should be more stable.

- One bug is that when `normed=True`, the errorbars have some problems (see run [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-08_2&test=Nz_r_DEEP2_JAN&right=2018-06/2018-06-08_2/Nz_r_DEEP2_JAN/config.yaml) with the above configuration issues fixed).  I suspect the histogram normalization scheme may be inconsistent between the jackknife calculation and the overall histogram, which would cause the covariance to be misestimated.

- Another bug is that when `normed=False`, the data and sims have inconsistent histogram normalization schemes, so they cannot be compared.  You can see this in another run [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-08_3&test=Nz_r_DEEP2_JAN&right=2018-06/2018-06-08_3/Nz_r_DEEP2_JAN/config.yaml).  However, note that in this case the errorbars on the simulation histogram actually seem a lot more reasonable!  So this justifies the conclusion above that the `normed=True` plots have a bug in the errorbar calculation.

yymao:
This PR fixes #117 by doing the normalization after the jackknife errors are calculated. It also updates the config files to use the correct argument label, and updates the number of bins and number of jackknife region to be 10 and 20, respectively, so that the jackknife error is meaningful. 

[Here's an example run using this PR, setting `normed=True` and using jackknife](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-09_3&test=Nz_r_DEEP2_JAN).

On the other hand, when `normed` is set to False, the test is not particularly meaningful because the validation data is normalized anyway, and different catalogs have different normalizations so the comparison is not really meaningful. [Here's a run with `normed=False`](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-09_4&test=Nz_r_DEEP2_JAN). The test is doing what it supposes to do (not normalize the datasets), but I don't think this particular setting is useful.


descqa/NumberDensityVersusRedshift.py
```diff
@@ -58,9 +58,9 @@ class NumberDensityVersusRedshift(BaseValidationTest):
     validation_color = 'black'
     default_markers = ['o', 'v', 's', 'd', 'H', '^', 'D', 'h', '<', '>', '.']
 
-    def __init__(self, z='redshift_true', band='i', N_zbins=44, zlo=0., zhi=1.1,
+    def __init__(self, z='redshift_true', band='i', N_zbins=10, zlo=0., zhi=1.1,
                  observation='', mag_lo=27, mag_hi=18, ncolumns=2, normed=True,
-                 jackknife=False, N_jack=10, ra='ra', dec='dec', pass_limit=2., use_diagonal_only=True,
+                 jackknife=False, N_jack=20, ra='ra', dec='dec', pass_limit=2., use_diagonal_only=True,
```
rmandelb: Good to change the defaults to be more reasonable.  I would suggest adding a comment indicating to users what are the considerations for the choice of `N_zbins` and `N_jack`.

yymao: added docstring

yymao:
@evevkovacs while I was adding the docstring, I wonder why we would ever want to set `use_diagonal_only` to be `True`? Shouldn't we always use the full covariance matrix to calculate chi^2?

evevkovacs:
@yymao @rmandelb  @janewman-pitt-edu  With regard to the normed=False option, that was intended to be used in combination with the option that doesn't plot any validation data, in case people want to see the actual numbers in each bin.  For example, for g band, there is no validation data, and one could plot the catalog data normalized or un-normalized. I thought that option is worth keeping, but if neither of you think so, then sure, we can take it out.  In any case, I agree that documentation was lacking.

yymao:
@evevkovacs I've added docstring. I wasn't planning on removing `normed` but just make it clear what it does in the docstring

evevkovacs:
@yymao  @rmandelb When I was first setting up the jackknife errors, Patricia cautioned me that when the statitsics were low, the off-diagonal elements of the covariance matrix could be flaky. Indeed in some of my tests I had problems with this, which is why I invoked that setting. But maybe this was caused by the bug in the code? if tests are now working with the full matrix, then it is fine to set that option to False.

evevkovacs:
@yymao Thanks Yao, I should have done a better job of filling those in. I have a better idea of the accepted format now.

yymao:
@evevkovacs (cc @rmandelb), OK, so the chi^2 values are very different when `use_diagonal_only` is on and off. See an example here https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-11 the test with `_d` postfix has `use_diagonal_only` set to `True`. 

Is this due to the off-diagonal terms of the covariance matrix being not stable? The difference is so large that I wonder if there is something else is wrong?

evevkovacs:
@yymao @rmandelb Let me take a look. I agree that the difference between the options seems implausible.

rmandelb:
I can think of three reasons for a difference in chi2 in these options:

- The off-diagonal covariances are very large (i.e., very high correlation coefficients between bins).  We do expect some non-negligible covariances in this case.

- Noise in the covariance matrices.  In general we know that jackknife covariances are noisy, so their inverse is a biased estimator of the true inverse covariance (which goes into the chi^2).  This will tend to inflate the chi^2 values, more so when using the full covariance than when using just the diagonal elements.  The standard correction for this, described in https://arxiv.org/pdf/astro-ph/0608064.pdf, would in this case (given the number of z bins and jackknife samples) reduce the chi^2 by a factor of ~0.5, which isn't enough to account for what we're seeing.

- A bug :)

janewman-pitt-edu:
I tend to suspect Rachel's second option.  20 jacknife samples for 10 parameters is going to yield a very very noisy covariance matrix.  How large an area of sky are these based on?

Also note that there are certainly uncertainties in our knowledge from DEEP2.  You should probably ascribe a ~20% error bar to those curves in calculating chi-squared.

janewman-pitt-edu:
By the way, just take a look at C^{-1} and that should show you pretty readily if the inversion was unstable...

yymao:
The chi^2 calculation is done correctly. Next thing is to print out the covariance matrix to inspect.

evevkovacs:
@janewman-pitt-edu @rmandelb @yymao Indeed, I believe it was an unstable inversion in the old code that led me to introduce the use_diagonal_only option.

yymao:
@evevkovacs thanks. Do you think the test should print out the covariance matrix? If so, do you want to take a stab to implement that? I can also do that but won't have time till later this week.

evevkovacs:
@yymao  Yes, I think that would be a useful addition. I will clone the branch and take a look at it.

janewman-pitt-edu:
I think a visualization (as opposed to printout) of the *inverse* covariance matrix would be most useful.

evevkovacs:
@janewman-pitt-edu OK. I may start with a printout because it is easier and then we can see if more is needed.

janewman-pitt-edu:
Just a note, by the way -- if you treat the 20% uncertainty on the DEEP2 n(z) as a diagonal error, including it before inverting should help stabilize the inversion (along the lines of ridge regression).

Really it shouldn't be diagonal -- errors in the fit will be covariant from bin to bin -- but that wouldn't help the inversion...

evevkovacs:
@yymao  @janewman-pitt-edu  Please look [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-14_7&catalog=proto-dc2_v4.6.1&right=2018-06/2018-06-14_7/Nz_i_Coil2004_maglim/proto-dc2_v4.6.1/Nvsz_LSST_i.txt) The coefficients of the inverse matrix are printed out at the end of the text file (just scroll down).

janewman-pitt-edu:
I think I'd be comfortable calling values like +/- 1E6 (when the typical value is in the hundreds) unstable.

yymao:
OK, I've added `use_diagonal_only: True` in the config files when jackknife is turned on, as it seems that the full covariance is not stable. We could try @janewman-pitt-edu's suggestion, but I think that should be a separate issue/PR. Since this PR is more about a bug fix, I'd like to merge it asap now that the bug (about normalization) in consideration is fixed. 

@rmandelb Can you update your review when you have a chance, thanks?

(cc @evevkovacs)

PR #120, opened 2018-06-18
changed shear label to shear_2_phosim, updated theory

### LINKED ISSUE #116: update sign convention and validation criteria in shear-shear test
@patricialarsen (cc @EiffL) following the update to the  shear sign convention in the GCRCatalogs (see 
https://github.com/LSSTDESC/gcr-catalogs/pull/111) we should make sure the shear-related tests also follow the same convention (that is, the treecorr/GalSim convention). 

We need to, for example, remove the minus sign in 
https://github.com/LSSTDESC/descqa/blob/master/descqa/shear_test.py#L172
https://github.com/LSSTDESC/descqa/blob/master/descqa/shear_test.py#L273 
(and maybe other places; I haven't done an extensive check).

patricialarsen:

Simple extension, changing the label for "shear_2" to "shear_2_phosim" and other minor changes in theory code. Note that to run this the GCR now needs to be up-to-date. Test result for v4.6 below. 

https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-18_6&test=shear&catalog=proto-dc2_v4.6.1

[edited by @yymao: fixes #116]


patricialarsen:
@yymao this is because I don't have the covariance matrices yet so the chi2 check is a very basic order of magnitude check on the ++ signal. The output value is still meaningful but it doesn't take into account the -- signal. I can change it to an "inspect" result in the meantime if you prefer.

yymao:
@patricialarsen yep, I think before there's meaningful chi^2 check we should use `inspect=True` so that people will look at the plots to make sure everything is fine. Thanks!

patricialarsen:
Okay I'm just testing it then I'll update the pull request - now if it fails the order of magnitude test it outright fails, otherwise it gets listed as inspect rather than pass.

patricialarsen:
New result if shears aren't immediately flagged as bad:
https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-19&test=shear&catalog=proto-dc2_v4.6.1_test

yymao:
@patricialarsen, thanks for the update. Running the latest version with the wrong `shear_2` conversion, I got an "inspect" result --- if this is intended that's OK. Since you mentioned above that if the result differs from the theory by an order of magnitude it would fail the test, I thought I'd check with you. In the case when `shear_2` convention is wrong, the result does differ from the theory by an order of magnitude:

https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-19_1&test=shear_wrong&right=2018-06/2018-06-19_1/shear_wrong/proto-dc2_v4.7_test/plot.png

patricialarsen:
@yymao  that's fine, I meant when the chi+ result varies by an order of magnitude not the chi-. Now it'll only return inspect or fail rather than pass so should be okay regardless.

yymao:
OK, thanks @patricialarsen. It's good then. I'll merge this by the end of tomorrow. I'm cc'ing @EiffL and @rmandelb in case they have further comments.

PR #122, opened 2018-06-26
Update color cut in CLF test

### LINKED ISSUE #121: color cut in CLF test should work on catalogs that have redshift blocks
[This issue](https://github.com/LSSTDESC/descqa/issues/50#issuecomment-399610103) that @evevkovacs encountered was due to the fact that the color cut implementation in the CLF test assumes the catalog does not iterate over redshift blocks. When a catalog iterates over redshift blocks, some blocks may not have the redshift ranges needed to determine the color cut, and hence results in error. 

This can be fixed if the test first determine the color cut using all blocks.

yymao:
This PR fixes #121. Here's an example run: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-26_9&test=CLF_r 

However, @evevkovacs, I wasn't able to test this on the new UM catalog. Can you point me a working GCRCatalogs that has the UM catalog in the configs? Thanks!

### DIFF
--- a/descqa/clf_test.py
+++ b/descqa/clf_test.py
@@ -18,6 +18,7 @@ class ConditionalLuminosityFunction(BaseValidationTest):
         self.mass_bins = 10**np.array((kwargs.get('mass_bins', [13.5, 14.0, 14.5, 15.0, 15.5])))
         self.z_bins = np.linspace(*kwargs.get('z_bins', (0.2, 1.0, 4)))
         self.color_cut_fraction = float(kwargs.get('color_cut_fraction', 0.2))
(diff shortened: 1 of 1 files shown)


evevkovacs:
@yymao  Yes, please use -p /global/u1/k/kovacs/gcr-catalogs_um  The catalog name is um_hpx_v0.0 and now covers a few hundred sq. deg. If you need something smaller (eg 1 healpix) I just made a config file for um_hpx_test_368, but haven't tested it yet. Let me know if you get errors.

yymao:
@evevkovacs thanks, that works. Here is the DESCQA run of this PR on the UM catalog: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-06-26_15&test=CLF_r

PR #123, opened 2018-06-29
DM output validation

fjaviersanchez:
With this PR I am creating a test that matches two catalogs and plots the difference in magnitude and position. @yymao I tried to follow your advice and set up the tests in `conclude_test`. Right now, I am reading the magnitudes from `mag_true_$BAND_lsst` but I think it would be useful to use something like `mag_$BAND` as an alias for the magnitudes that we want to compare.

I am using the simple `KDTree` nearest neighbor matcher (`spatial_closest`) from `https://github.com/LSSTDESC/CatalogMatcher` for speed purposes but I am open to use any other matching technique. 

For DM catalogs we also need to set up the selection in the config file (and this selection cuts should be transparent somehow to the truth catalogs). Right now, I have as a placeholder `ra < 500`. 

So, as to-do list:

- Generalize magnitude reading.
- Generalize selection.


yymao:
Oh i see you closed this one already.

fjaviersanchez:
I started an new branch but for some reason it branched out from `bias_test` and not from `master`, sorry about that! I'll try to clean up the stuff and make a new PR :)

PR #124, opened 2018-06-29
DM catalog validation

fjaviersanchez:
With this PR I am creating a test that matches two catalogs and plots the difference in magnitude and position. @yymao I tried to follow your advice and set up the tests in conclude_test. Right now, I am reading the magnitudes from mag_true_$BAND_lsst but I think it would be useful to use something like mag_$BAND as an alias for the magnitudes that we want to compare.

I am using the simple KDTree nearest neighbor matcher (spatial_closest) from https://github.com/LSSTDESC/CatalogMatcher for speed purposes but I am open to use any other matching technique.

For DM catalogs we also need to set up the selection in the config file (and this selection cuts should be transparent somehow to the truth catalogs). Right now, I have as a placeholder ra < 500.

So, as to-do list:

Generalize magnitude reading.
Generalize selection.

### DIFF
--- a/descqa/CheckAstroPhoto.py
+++ b/descqa/CheckAstroPhoto.py
@@ -1,0 +1,125 @@
+from __future__ import print_function, division
+import os
+import numpy as np
+from scipy.stats import binned_statistic
+from CatalogMatcher.match import spatial_closest  # https://github.com/LSSTDESC/CatalogMatcher
+from GCR import GCRQuery
+from .base import BaseValidationTest, TestResult
+from .plotting import plt
+from matplotlib.ticker import NullFormatter
+
+__all__ = ['CheckAstroPhoto']
+
+nullfmt = NullFormatter()
+
+class CheckAstroPhoto(BaseValidationTest):
+    """
+    Validation test to compare astrometric and photometric results between
+    two different datasets.
+    """
+    
+    def __init__(self, **kwargs):
+        #pylint: disable=W0231
+        self.kwargs = kwargs
+        self.min_mag = kwargs['min_mag']  # Minimum magnitude to bin the sample
+        self.max_mag = kwargs['max_mag']  # Maximum magnitude to bin the sample
+        self.nbins = kwargs['nbins_mag'] # Number of bins
+        self.ra=dict() # Here we are going to store the RA for all catalogs
+        self.dec=dict() # Here we are going to store the DEC for all catalogs
+        self.magnitude=dict() # Here we are going to store the magnitude (in different bands for all catalogs)
+        self.selection_cuts = kwargs['selection_cuts'] # Selection cuts to perform on the data sample
+        self.bands = kwargs['bands'] # Photometric band(s) to analyze
+      
+    def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
+
+        #TODO: Use aliases for magnitudes and make this more general
+        mags = ['mag_true_%s_lsst' % band for band in self.bands]
+        qs = ['ra', 'dec']
+        qs.append(mags)
+        filters = [GCRQuery(self.selection_cuts)]
+        data = catalog_instance.get_quantities(['ra','dec','mag_true_r_lsst'], filters=filters)
+        print('Selected %d objects for catalog %s' % (len(data), catalog_name))
+        self.ra[catalog_name] = data['ra']
+        self.dec[catalog_name] = data['dec']
+        
+        for band in self.bands:
+            self.magnitude[(catalog_name, band)] = data['mag_true_%s_lsst' % band]
+        return TestResult(inspect_only=True)
+
+    def scatter_project(self, x, y, xmin, xmax, ymin, ymax, nbins, xlabel, ylabel, savename, bin_stat=False):
+
+        left, width = 0.1, 0.65
+        bottom, height = 0.1, 0.65
+        bottom_h = left_h = left + width + 0.02
+
+        rect_scatter = [left, bottom, width, height]
+        rect_histx = [left, bottom_h, width, 0.2]
+        rect_histy = [left_h, bottom, 0.2, height]
+
+        plt.figure()
+
+        axScatter = plt.axes(rect_scatter)
+        axHistx = plt.axes(rect_histx)
+        axHisty = plt.axes(rect_histy)
+
+        axHistx.xaxis.set_major_formatter(nullfmt)
+        axHisty.yaxis.set_major_formatter(nullfmt)
+
+        axScatter.scatter(x, y, s=1., alpha=0.5)
+        
+        if bin_stat:
+            mean_y, be, _ = binned_statistic(x, y, range=(xmin, xmax), bins=nbins)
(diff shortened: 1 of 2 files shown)


fjaviersanchez:
An example of the plots that this test generates (using `proto-dc2_v2.0_test` and `proto-dc2_v3.0_test`)

screen shot 2018-06-29 at 12 41 06 pm
screen shot 2018-06-29 at 12 40 47 pm

(Since the galaxies in these catalogs are unrelated we can expect that we do not find any correlation between positions nor magnitudes).

fjaviersanchez:
Success! Thanks @yymao:

screen shot 2018-07-27 at 11 13 10 am
screen shot 2018-07-27 at 11 12 59 am

rmandelb:
@fjaviersanchez - do you understand why your red points in the cmodel mag plot are offset by ~0.1-0.2 magnitudes from the bright part of the 2D histogram (which appears to be centered at Delta mag~0)?

Could there be a few outliers that are throwing the mean off, in which case the median (or a mean determined with outlier rejection) would be more stable?

fjaviersanchez:
@rmandelb I think those are bad matches. I am only using nearest-neighbor matching (because it's the fastest) and I think that some sources are being wrongly associated to a fainter source creating pretty big outliers. I am including the median here:

image

I think that I'll include both mean and median (the more information the better I guess).

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def conclude_test(self, output_dir):
+        """
+        This function should gather the two catalogs, match them and perform the summary plots
+        """
+
+        try:
+            assert(len(self.ra.keys())==2) # Making sure that we have *just* two catalogs
+        except AssertionError:
+            print('The test can compare two catalogs only!')
```
yymao: raise an error if the test shouldn't work
```python
        if len(self.ra) != 2: # Making sure that we have *just* two catalogs
            raise ValueError('The test can compare two catalogs only!')
```

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+
+        cat_names = list(self.ra.keys()) # This is an auxiliary list to easily get the catalogs
```
yymao: unnecessary `.keys()`
```python
cat_names = list(self.ra)
```

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+        print(cat_names)
+        cat_len = [len(self.ra[cat_names[0]]),len(self.ra[cat_names[1]])]
+        cat_names = np.array(cat_names)[np.argsort(cat_len)].tolist() # We find the catalog with less objects to build the tree
```
yymao: Lines 104-105, simply the logic
```python
cat_names = sorted(cat_names, key=lambda name: len(self.ra[name]))
```

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+        
+        # Scatter plot + histogram of Delta mag vs mag
+        for band in self.bands:
+            photo_savename = os.path.join(output_dir, 'photometry_check_%s_%s_%s.png' % (cat_names[0], cat_names[1], band))
+            self.scatter_project(self.magnitude[(cat_names[0], band)][good_mag[band]],
+                delta_mag[band][good_mag[band]], self.min_mag, self.max_mag,
+                -1, 1, self.nbins, '%s' % band, r'$\Delta %s$' % band, photo_savename, bin_stat=True)
+            n_true, _ = np.histogram(self.magnitude[(cat_names[1], band)], bins=50, range=(10, 30))
+            n_meas, bin_edges = np.histogram(self.magnitude[(cat_names[0], band)], bins=50, range=(10, 30))
+            plt.figure()
```
yymao: Store the figure object so that we can close it later.
```python
fig = plt.figure()
```

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+            plt.plot(0.5*(bin_edges[1:]+bin_edges[:-1]), 1.0*n_meas/n_true,'o')
+            plt.xlabel('{}'.format(band))
+            plt.ylabel('Ratio of detected over input objects')
+            plt.ylim(0,1)
+            plt.tight_layout()
+            photo_savename = os.path.join(output_dir, 'mag_ratio_%s_%s_%s.png' % (cat_names[0], cat_names[1], band))    
+            plt.savefig(photo_savename)
```
yymao: Add after Line 139
```python
plt.close(fig)
```
so that matplotlib will not waste memory on open figures.

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+        # Scatter plot + histogram of Delta mag vs mag
+        for band in self.bands:
+            photo_savename = os.path.join(output_dir, 'photometry_check_%s_%s_%s.png' % (cat_names[0], cat_names[1], band))
+            self.scatter_project(self.magnitude[(cat_names[0], band)][good_mag[band]],
+                delta_mag[band][good_mag[band]], self.min_mag, self.max_mag,
+                -1, 1, self.nbins, '%s' % band, r'$\Delta %s$' % band, photo_savename, bin_stat=True)
+            n_true, _ = np.histogram(self.magnitude[(cat_names[1], band)], bins=50, range=(10, 30))
+            n_meas, bin_edges = np.histogram(self.magnitude[(cat_names[0], band)], bins=50, range=(10, 30))
+            plt.figure()
+            plt.plot(0.5*(bin_edges[1:]+bin_edges[:-1]), 1.0*n_meas/n_true,'o')
```
yymao: Instead of `1.0*n_meas/n_true`, do `n_meas.astype(float) / n_true` to be explicit what the code does.

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+
+        left, width = 0.1, 0.65
+        bottom, height = 0.1, 0.65
+        bottom_h = left_h = left + width + 0.02
+
+        rect_scatter = [left, bottom, width, height]
+        rect_histx = [left, bottom_h, width, 0.2]
+        rect_histy = [left_h, bottom, 0.2, height]
+
+        plt.figure()
```
yymao: store the figure object so that we can close it later
```
fig = plt.figure()
```

descqa/CheckAstroPhoto.py
```diff
@@ -0,0 +1,139 @@
...
+        axScatter.set_ylabel(ylabel)
+
+        axHistx.hist(x, bins=nbins, range=(xmin, xmax))
+        axHisty.hist(y, bins=nbins, range=(ymin, ymax), orientation='horizontal')
+
+        axHistx.autoscale(tight=True)
+        axHisty.autoscale(tight=True)
+        axHistx.set_xlim(axScatter.get_xlim())
+        axHisty.set_ylim(axScatter.get_ylim())
+        plt.savefig(savename)
```
yymao: add after Line 90
```python
plt.close(fig)
```

Issue #134, opened 2018-08-08
Density vs extinction test


fjaviersanchez:
This issue is to track the progress on the implementation of the test that checks the density as a function of the MW extinction. In principle, the test will be most relevant for the DC2 object catalogs. However, in principle, it can be applied to any  catalog.

yymao:
Closed by #140

PR #140, opened 2018-08-15
First commit of Density test

fjaviersanchez:
This PR tries to cover one of the tests that are in #123 and solves #134. I am selecting the galaxies from the DM object catalogs using their `extendedness` values  (in particular I require `extendedness==1`). I have an example here: https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-08-14_10&test=DensityVersusExtinction&right=2018-08/2018-08-14_10/DensityVersusExtinction/dc2_coadd_run1.1p/dc2_coadd_run1.1p_density_vs_extinction.png 

@rmjarvis Right now everything is very vanilla and I assumed that there's no depth variation but no cuts have been made to ensure this. Should I include something to do this?


descqa/DensityVersusExtinction.py
```diff
@@ -0,0 +1,51 @@
...
+        self.validation_path = kwargs['validation_map_filename']
+        self.nside = kwargs['nside'] 
+        self.validation_data = hp.ud_grade(hp.read_map(self.validation_path), nside_out=self.nside)
+        self.xlabel = kwargs['xlabel']
+     
+    def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
+        
+        catalog_data = catalog_instance.get_quantities(['ra', 'dec', 'extendedness'])
+        galaxies = catalog_data['extendedness']==1 
+        data_map = create_hp_map(catalog_data['ra'][galaxies], catalog_data['dec'][galaxies], self.nside)
```
yymao: Lines 34-36: (1) add `has_quantities` and skip test if not all needed quantities exist, (2) use `filters` instead of masking the data afterwards.
```python
        if not catalog_instance.has_quantities(['ra', 'dec', 'extendedness']):
            return TestResult(skipped=True, summary='catalog does not have needed quantities')
        catalog_data = catalog_instance.get_quantities(['ra', 'dec', 'extendedness'], filters=['extendedness == 1'])
        data_map = create_hp_map(catalog_data['ra'], catalog_data['dec'], self.nside)
```

descqa/DensityVersusExtinction.py
```diff
@@ -0,0 +1,51 @@
...
+        mask = data_map>0 # This is a good approximation if the pixels are big enough
+        xmin = np.percentile(self.validation_data[mask], 5)
+        xmax = np.percentile(self.validation_data[mask], 95)
+        data_map = data_map/(3600*hp.nside2pixarea(self.nside, degrees=True)) # To get the density in arcmin^-2
```
yymao: suggestion for Line 40
```python
data_map /= (3600*hp.nside2pixarea(self.nside, degrees=True))
```

descqa/DensityVersusExtinction.py
```diff
@@ -0,0 +1,51 @@
...
+        self.xlabel = kwargs['xlabel']
+     
+    def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
+        
+        catalog_data = catalog_instance.get_quantities(['ra', 'dec', 'extendedness'])
+        galaxies = catalog_data['extendedness']==1 
+        data_map = create_hp_map(catalog_data['ra'][galaxies], catalog_data['dec'][galaxies], self.nside)
+        mask = data_map>0 # This is a good approximation if the pixels are big enough
+        xmin = np.percentile(self.validation_data[mask], 5)
+        xmax = np.percentile(self.validation_data[mask], 95)
```
yymao: suggestion for Lines 38-39
```python
xmin, xmax = np.percentile(self.validation_data[mask], [5, 95])
```

descqa/DensityVersusExtinction.py
```diff
@@ -0,0 +1,51 @@
...
+import healpy as hp
+
+__all__ = ['DensityVersusExtinction']
+
+def create_hp_map(ra, dec, nside):
+    """
+    Auxiliary function to generate HEALPix maps from catalogs.
+    It reads the ra and dec in degrees and returns a HEALPix map
+    """
+    pixnums = hp.ang2pix(nside, np.pi/2-np.radians(dec), np.radians(ra))
```
yymao:
```python
    pixnums = hp.ang2pix(nside, ra, dec, lonlat=True)
````

File: descqa/DensityVersusExtinction.py [hunk 5]
```diff
@@ -0,0 +1,51 @@
...
+    return np.bincount(pixnums, minlength=hp.nside2npix(nside))
+
+class DensityVersusExtinction(BaseValidationTest):
```
yymao: Will this class better be named as `DensityVersusSkyPosition` or something along the line? In particular, since `validation_map_filename` and `xlabel` are both configurable options, one can supply another validation map that is not related to extinction, right? Then it seems to me that the class name is too restrictive?

fjaviersanchez: You are totally right. Thanks!

fjaviersanchez:
Thank you @yymao. Yes, let's wait for @rmjarvis to see whether he has some suggestions for additions/refinements please.

descqa/DensityVersusSkyPosition.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    pixnums = hp.ang2pix(nside, ra, dec, lonlat=True)
+    return np.bincount(pixnums, minlength=hp.nside2npix(nside)).astype(float)
+
+class DensityVersusSkyPosition(BaseValidationTest):
+    """
+    This tests checks the object density as a function
+    of another map (e.g: extinction, airmass, etc)
+    """
+    def __init__(self,**kwargs): # pylint: disable=W0231
```
rmjarvis: Why use `**kwargs` here rather than list the kwargs?  It looks like the are all required, so normally that would be good to list explicitly in the function declaration.  Also worth mentioning what the parameters are in the doc string.

fjaviersanchez: I did this to be able to "recycle" this script with a different configuration file (and to match the style of most of the tests in `descqa`). @yymao is there an easy way to use positional arguments and be able to read from different config files?

yymao: @fjaviersanchez I'm not quite sure what you mean "use positional arguments and be able to read from different config files." It is OK to use positional arguments in DESCQA tests; the only requirement is that you must allow additional keyword arguments at the end. See [this test](https://github.com/LSSTDESC/descqa/blob/master/descqa/NumberDensityVersusRedshift.py#L103) for an example. That being said, since these test classes are rarely being called by users in interactive environments, I'm ok with just using `**kwargs`. The more important thing is to specify the arguments in docstring which I see you already did.

PR #146, opened 2018-08-22
Adding validation test for stellar mass distribution of galaxies and their number densities

### LINKED ISSUE #109: Stellar mass distribution
The purpose of this issue is to understand the shape of the distribution of galaxy stellar masses in protoDC2 catalogues as we have been discussing this with @reneehlozek. In the previous version, there was a small bump at about 10^10 solar masses that was not obvious to me why it was there. In the most recent version, the bump is still there as well as another skewness at lower masses around 10^8 solar masses. Maybe this second bump is because that the peak of the distribution was about 10^8 solar masses in version 2 while the peak is currently at much lower masses around 10^(6-7) solar masses. The histograms are shown below:
mstellar

I checked the stellar mass distribution from SDSS galaxies from Maraston et al. 2013 that plotted the distributions for SDSS BOSS and CMASS galaxies that were fit with different templates. These distribution do not have any obvious bumps. The figures from the paper are shown below for BOSS and CMASS catalogues:
sdss_mstellar

I also know that the CMASS galaxies are more biased toward higher mass galaxies but it would be good to know why protoDC2 galaxies peak at such small stellar masses.

Andromedanita:
Validation test for issue #109 

[edited by @yymao: fix #109]


yymao:
@Andromedanita thanks very much for the PR! I guess you are still working on this? (which is totally OK -- we encourage early PR, just let me know when it is ready to be reviewed.)

I did spot a few things already: to import from descqa submodule, you'll need to add a `.` to your import statement: 
```python
from .base import BaseValidationTest, TestResult
from .plotting import plt
```

Also, remove these:
```python
import matplotlib.pylab as plt
%matplotlib inline
```

And, you'll need to add a yaml config file too.  Let me know if you have any questions!

yymao (changes requested):
I have edited the code so that it runs and passes basic QA. However, this PR is still missing a test config file, and also an actual DESCQA run.

Andromedanita:
@yymao I have edited the code by a tiny bit (changed the title of the plot and also includes cmass plot).
Should I commit that still?

Also, what is the step that I need to do to add a test config file and  actual DESCQA run that you mentioned?

yymao:
@Andromedanita yes, please commit. Just watch out for potential merge conflicts. 

To add a test config, just add a yaml file (example below) inside `descqa/configs` and give it a short name.

```yaml
subclass_name: StellarMassDistribution.StellarMassTest
description: <add a description here>
```

yymao:
@Andromedanita (cc @evevkovacs), I have updated this branch to make the test work within the DESCQA framework, and also utilize `GCRQuery` to be more efficient. I've also removed the redshift calculation because (1) it is not used at all in the test and (2) redshift is available in the cosmoDC2 catalog, and I don't see why one would need to recalculate it. 

@Andromedanita  Can you review my updates and see if I implemented your test correctly. If you want to make future changes, make sure you pull from this branch first before you add new changes. 

@evevkovacs  Can you test if this updated test now works?

Andromedanita:
@yymao I reviewed the code and it seems perfect! Thanks for reviewing it and making the necessary changes!

yymao:
I've just run this test on `cosmoDC2_v1.1.4_small`. Here's the test result: https://portal.nersc.gov/cfs/lsst/descqa/v2/?run=2021-02-10&test=StellarMassDistruibution&catalog=cosmoDC2_v1.1.4_small

image

yymao (approved):

PR #148, opened 2018-08-24
Changes for readiness and sky area tests to run on baseDC2 and cosmoDC2

evevkovacs:
Here are the links for these tests:
[readiness test](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-08-23_21&test=readiness_baseDC2)
The portal seems to be down so I can't link the area test, but it was in the DC2 slides for 8/24/18

### DIFF
--- a/descqa/readiness_test.py
+++ b/descqa/readiness_test.py
@@ -347,12 +347,8 @@ class CheckQuantities(BaseValidationTest):
             if not catalog_instance.has_quantities(quantities_needed):
                 self.record_result('{} does not exist'.format(' or '.join(quantities_needed)), failed=True)
                 continue
-            
-            try:
-                data = catalog_instance.get_quantities(quantities_needed)
-            except KeyError:
-                raise ValueError('{} not found'.format(', '.join(quantities_needed)))
 
+            data = catalog_instance.get_quantities(quantities_needed)
             if check_uniqueness(data[quantity], data.get(mask)):
                 self.record_result('{} is all unique'.format(label))
             else:
--- a/descqa/readiness_test.py
+++ b/descqa/readiness_test.py
@@ -138,8 +138,11 @@ class CheckQuantities(BaseValidationTest):
 
     def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
 
-        all_quantities = sorted(catalog_instance.list_all_quantities(True))
+        #list_of_all_quantities = sorted(map(str,catalog_instance.list_all_quantities(True)))
+        #using True in the previous statement means that the elements require further parsing to remove the catalog name for composite catalogs
+        #all_quantities = [re.sub("'", "", q.split(', ')[-1].split(')')[0]) for q in list_of_all_quantities]
 
+        all_quantities = sorted(catalog_instance.list_all_quantities())
         galaxy_count = None
         failed_count = 0
         output_rows = []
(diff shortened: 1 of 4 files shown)


descqa/readiness_test.py
```diff
@@ -138,8 +138,11 @@ def _format_row(self, quantity, plot_filename, results):
 
     def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
 
-        all_quantities = sorted(catalog_instance.list_all_quantities(True))
+        #list_of_all_quantities = sorted(map(str,catalog_instance.list_all_quantities(True)))
+        #using True in the previous statement means that the elements require further parsing to remove the catalog name for composite catalogs
+        #all_quantities = [re.sub("'", "", q.split(', ')[-1].split(')')[0]) for q in list_of_all_quantities]
```
yymao: remove unused code/comments

descqa/readiness_test.py
```diff
@@ -138,8 +138,11 @@ def _format_row(self, quantity, plot_filename, results):
...
 
+        all_quantities = sorted(catalog_instance.list_all_quantities())
```
yymao: This change makes the test not backward compatible. There are cases we would want the readiness test to run on all quantities, including native ones. I think the way to go is to add a test option like `exclude_native`, and we can turn it on for composite catalogs.

descqa/readiness_test.py
```diff
@@ -267,8 +270,12 @@ def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
                 output_header.append('<span class="fail">{} does not exist!</span>'.format(' or '.join(quantities_needed)))
                 failed_count += 1
                 continue
+            
+            try:
+                data = catalog_instance.get_quantities(quantities_needed)
+            except KeyError:
+                raise ValueError('{} not found'.format(', '.join(quantities_needed)))
```
yymao: (Lines 273-277) this seems to be a hack that is no longer needed.

descqa/readiness_test.py
evevkovacs: So the commented out lines above (141-143) are what is needed if one sets True in the list_all_quantities call. Perhaps you know a more elegant way to do this but the idea is to parse the string to see if it came from a tuple (hence split on ', '), take the last argument of that split, which is the desired quantity, and remove the trailing bracket and any internal quotes. I have tested that those lines work.  I'll double check, but with those additions the test will continue to be backwards compatible and will work on both native and non-native quantities in composite catalogs.

yymao: Generally, the test should not assume that the quantity names are strings, nor assume that the quantity names are tuples. The readiness test does assume the quantity names are strings. So what we can do here is (1) add an option to toggle all or just GCR quantities and/or (2) remove any quantities that have non-string names.

descqa/readiness_test.py
evevkovacs: OK, but since we know that the quantity names will be tuples in the case of composite catalogs why can't we put code in to handle that?

yymao: For composite catalogs, the readiness test should really only check derived quantities. The native quantities in a composite catalog are only used for accessing underlying catalogs individually, which is not what the readiness test trying to do (imagine that if the add-on fixes a bug in the main catalog, then the readiness test only needs to run on the fixed column, not the original column). Hence, all the tuple quantity names are not actually what the readiness test cares about.

PR #168, opened 2019-04-24
color distribution for snapshot catalogs

evevkovacs:
Here are the code modifications so that the Color distribution test will run on the baseDC2 snapshots (uses the new snapshot reader) Here is a test result: ( https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-04-23&catalog=baseDC2_snapshot_z0.15_v0.1)
Yao, I accidentally committed the same code to master, but it didn't get pushed. Sorry, I'm not sure how to clean that up.


evevkovacs:
Superseded by #186

PR #177, opened 2019-08-08
Modifications to run catalogs with or without agn contributions

evevkovacs:
Color distribution test required minor changes to be able to run on the new AGN catalog. Added a flag to select magnitudes with no agn contributions. Added lsst magnitudes as an option in case SDSS or DES magnitudes are unavailable. If LSST magnitudes used, plot title indicates this. Test results are [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-08-07_1 ). Test results for default case with AGN magnitudes included are [here](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-08-07_2)

### DIFF
--- a/descqa/ColorDistribution.py
+++ b/descqa/ColorDistribution.py
@@ -139,15 +139,16 @@ class ColorDistribution(BaseValidationTest):
         
         # Color transformation
         color_trans = None
-        if self.color_transformation_q and filter_this != 'lsst':
+        if self.color_transformation_q:
             color_trans_name = None
-            if self.validation_catalog == 'DEEP2':
+            if self.validation_catalog == 'DEEP2' and filter_this != 'lsst':
                 color_trans_name = '{}2cfht'.format(filter_this)
             elif self.validation_catalog == 'SDSS' and filter_this == 'des':
                 color_trans_name = 'des2sdss'
             if color_trans_name:
                 color_trans = color_transformation[color_trans_name]
 
+        filter_title = '\mathrm{{{}}}'.format(filter_this.upper())
         if color_trans:
             data_transformed = {}
             for band in bands:
@@ -156,10 +157,12 @@ class ColorDistribution(BaseValidationTest):
                 except KeyError:
                     continue
 
+            filter_title = '{}\\rightarrow\mathrm{{{}}}'.format(filter_title,
+                                                                   self.validation_catalog) if data_transformed else filter_title
             data_transformed['redshift'] = data['redshift']
             data = data_transformed
             del data_transformed
-
+            
         if self.obs_r_mag_limit and not self.rest_frame:
             data = GCRQuery('r < {}'.format(self.obs_r_mag_limit)).filter(data)
         elif self.Mag_r_limit and self.rest_frame:
@@ -187,7 +190,7 @@ class ColorDistribution(BaseValidationTest):
                     self.obs_color_dist[color]['binctr'], self.obs_color_dist[color]['cdf'])
 
         self.make_plots(mock_color_dist, color_shift, cvm_omega, cvm_omega_shift, catalog_name,
-                        output_dir, filter_this)
+                        output_dir, filter_title)
 
         # Write to summary file
         fn = os.path.join(output_dir, self.summary_output_file)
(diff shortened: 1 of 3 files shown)


evevkovacs:
@yymao travis checks passed. Please review, thanks!

descqa/ColorDistribution.py
```diff
@@ -60,7 +60,8 @@ def __init__(self, **kwargs): # pylint: disable=W0231
         self.color_transformation_q = kwargs.get('color_transformation_q', True)
         self.Mag_r_limit = kwargs.get('Mag_r_limit', None)
         self.rest_frame = kwargs.get('rest_frame', bool(self.Mag_r_limit and not self.obs_r_mag_limit))
-
+        self.use_agn = kwargs.get('use_agn', True)
```
yymao: Setting default value to `True` can cause confusion, as users usually expect that, when `use_agn` is not set, it means not using AGN. 

A better way would be to rename this argument to `exclude_agn`, which then can be default to `False`.

descqa/ColorDistribution.py
```diff
@@ -130,10 +136,11 @@ def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
                    '{} < {}'.format(labels['redshift'], self.zhi)]
         data = catalog_instance.get_quantities(list(labels.values()), filters)
         data = {k: data[v] for k, v in labels.items()}
-
+        print(list(data.keys()))
```
yymao: Is this print out statement needed?

descqa/ColorDistribution.py
```diff
@@ -111,12 +112,17 @@ def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
         if self.rest_frame:
             possible_names = ('Mag_{}_lsst', 'Mag_{}_sdss', 'Mag_true_{}_lsst_z0', 'Mag_true_{}_sdss_z0')
         else:
-            possible_names = ('mag_{}_sdss', 'mag_{}_des', 'mag_true_{}_sdss', 'mag_true_{}_des')
+            possible_lsst_names = ('mag_{}_lsst', 'mag_true_{}_lsst') if self.use_agn else ('mag_{}_noagn_lsst',
```
yymao: change to `exclude_agn`?

descqa/ColorDistribution.py
```diff
@@ -130,10 +136,11 @@ def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
                    '{} < {}'.format(labels['redshift'], self.zhi)]
         data = catalog_instance.get_quantities(list(labels.values()), filters)
         data = {k: data[v] for k, v in labels.items()}
-
+        print(list(data.keys()))
+        
         # Color transformation
         color_trans = None
-        if self.color_transformation_q:
+        if self.color_transformation_q and filter_this != 'lsst':
```
yymao: I don't think this second condition is needed, is it?  For filter that does not have corresponding transformation specified, transformation won't be applied.
```suggestion
        if self.color_transformation_q:
```

descqa/ColorDistribution.py
```diff
@@ -217,8 +226,8 @@ def make_plots(self, mock_color_dist, color_shift, cvm_omega, cvm_omega_shift, c
             title = '$m_r < {:2.1f},  {:.1f} < z < {:.1f}$'.format(self.obs_r_mag_limit, self.zlo, self.zhi)
         elif self.Mag_r_limit:
             title = '$M_r < {:2.1f},  {:.1f} < z < {:.1f}$'.format(self.Mag_r_limit, self.zlo, self.zhi)
-
-        print(mock_color_dist.keys())
+        title = title + ' (LSST filters)' if filters == 'lsst' else title
```
yymao: why not just always print out filters? 
```suggestion
        title += ' ({})'.format(filters.upper())
```

descqa/ColorDistribution.py
evevkovacs: I changed this as you suggested

descqa/ColorDistribution.py
evevkovacs: Removed.

descqa/ColorDistribution.py
evevkovacs: I moved this check 2 lines down to make sure that the color transformation for DEEP2 data is not called in the case that LSST filters are being used (since this trasnformation doesn't exist)

descqa/ColorDistribution.py
evevkovacs: This would be misleading in the case that the filters have undergone a color transformation. I added some code to include information about a filter transformation, if it occurred. Here's the example [DESCQA test](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-08-09_9) with the revised code.

evevkovacs:
@yymao I think the test is now ready and the plot title is improved. Please review.

Issue #173, opened 2019-07-08
`sklearn.cluster.k_means` not working when `n_jobs=-1` is set


yymao:
Potentially due to an update of `loky`, `sklearn.cluster.k_means` is not working in the current DESC Python environment when `n_jobs=-1` is set. This option should be unset. 

Thanks to @evevkovacs for discovering this issue.

yymao:
Closed by #179

PR #179, opened 2019-09-10
Fix issues with change to python 3 sklearn; add cosmetics to N vs z test

evevkovacs:
This PR addresses issue #173 to fix the bug caused by the update to sklearn in python 3. 
The call to k_means clustering changed. The shear test required another minor fix. Some cosmetic changes to the number-density test are also included.  DESCQA tests have been run in the past but I can't post links at the moment because nersc is still down. Will add these when cori is back up.

### DIFF
--- a/descqa/shear_test.py
+++ b/descqa/shear_test.py
@@ -164,7 +164,7 @@ class ShearTest(BaseValidationTest):
         N_clust = self.N_clust
         nn = np.stack((catalog_data[self.ra][mask], catalog_data[self.dec][mask]), axis=1)
         _, labs, _ = k_means(
-            n_clusters=N_clust, random_state=0, X=nn, n_jobs=-1)  # check random state, n_jobs is in debugging mode
+            n_clusters=N_clust, random_state=0, X=nn)  # check random state
         print("computing jack-knife errors")
         time_jack = time.time()
         # jack-knife code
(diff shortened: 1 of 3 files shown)


evevkovacs:
[Number-density test]https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-09-13&test=Nz_r_Coil2004_maglim()

evevkovacs:
And after resolving the import camb issue, here is the [shear test](https://portal.nersc.gov/project/lsst/descqa/v2/?run=2019-09-17_2). I think the code is ready to merge

descqa/shear_test.py
```diff
@@ -273,7 +273,7 @@ def run_on_single_catalog(self, catalog_instance, catalog_name, output_dir):
 
         pars.set_cosmology(H0=cosmo.H0.value, ombh2=cosmo.Ob0*cosmo.h**2, omch2=(cosmo.Om0-cosmo.Ob0)*cosmo.h**2)
         pars.InitPower.set_params(ns=ns, As=2.168e-9*(s8/0.8)**2)
-        camb.set_halofit_version(version='takahashi') # pylint: disable=no-member
+        #camb.set_halofit_version(version='takahashi') # pylint: disable=no-member
```
yymao: Is this line no longer needed? If so, remove it instead of commenting it out (because the residual comment makes people wonder why this is commented out and may cause future confusion. 
```suggestion
```

evevkovacs: @patricialarsen suggested this when I was testing for another problem. I defer to her on whether this line should be reinstated or removed.

patricialarsen: So the change is from an update to camb. IIRC the syntax is appropriate for the older version but will cause a crash in the new version. 

Let me double check what the default nonlinear model is for the matter power interpolator in case we need to replace this with the updated syntax.

evevkovacs:
These issues have been addressed in other updates

PR #188, opened 2019-11-20
Move CI test to GitHub action

yymao:
This PR moves the current travis-CI test to GitHub action because it runs faster, integrates with GitHub better, and has a cleaner output (so that it is easier to see what need to be fixed). 

With this PR I also fix a few existing linting issues.


yymao:
@evevkovacs thanks for the feedback, I've added an ["Integration test" section to `CONTRIBUTING.md`](https://github.com/LSSTDESC/descqa/blob/u/yymao/github-action/CONTRIBUTING.md#integration-test).

See if that works.

evevkovacs:
@yymao Thanks, you can merge, or let me know if you would like me to go ahead with it.

PR #191, opened 2019-11-22
EmissionlineRatioTest Fixes

adam-broussard:
The previous pull request had a commit in it (7a3acbf7997000af98494199b19e12c05d595c16) that undid the work of a previous commit.  This PR acts to correct the faulty commit.


adam-broussard:
I tried to do something clever here by modifying a past commit, but I'm realizing that I think it would just make the whole issue worse... I'm going to fix the previous merge a different way and open a different pull request...  Sorry for the email spam.

PR #194, opened 2020-01-21
Emlineqa Redshift Distribution Matching

adam-broussard:
Adds redshift distribution matching to emission line luminosity ratio test so that the DC2 data are accurately compared against the SDSS data.

[Web Interface](https://portal.nersc.gov/projecta/lsst/descqa/v2/?run=2020-01-21_6)


adam-broussard:
The linter appears to be crashing every time I run something - can anybody comment on this?

descqa/EmlineRatioTest.py
```diff
@@ -5,6 +5,7 @@
 # W0201 gives a warning when defining attributes outside of __init__()
 from __future__ import unicode_literals, absolute_import, division
 import os
```
yymao: Linter is complaining about unused import
```suggestion
```

adam-broussard:
Apologies - the linter was exiting with an error code, so I assumed that meant it wasn't running correctly, rather than that it had found something that needed to be modified!

PR #202, opened 2020-05-04
Update linter arguments in CI

yymao:
This PR updates the linter arguments in the CI test to avoid CI test failures due to unnecessary linter complaints.


yymao:
Thanks, @evevkovacs. 

(`v2` there is refering to the version of the "checkout" action, not DESCQA.)

PR #206, opened 2020-10-13
Update CGI python executable path

yymao:
This PR updates the CGI python executable path to use the current DESC python environment.


yymao:
PR updated with the new environment as described in https://github.com/LSSTDESC/desc-help/issues/45. 

@heather999 can you review this PR? Thanks.

heather999 (approved):
That's the right path.  Double-checked it at NERSC - looks good!

yymao:
Thanks, @heather999 .

PR #208, opened 2020-12-04
Added config to filter for halo_id

fjaviersanchez:
Added configuration file for mock specific `bias_test` (selects objects with positive `halo_id`) and modified name of previous test (which will work with object catalogs with redshift and mocks but may include unclustered objects).


evevkovacs (approved):
Thanks very much. I think the maximum limit on halo_id is fine for the forseeable future because we won't get a catalog that is larger than the number you give in the configuration file.

fjaviersanchez:
Thank you! Should I merge?

evevkovacs:
Thanks I just merged

PR #210, opened 2020-12-04
Remove module versions

### LINKED ISSUE #209: Need to update versions of gsl and cray-fftw in run_master.sh
See [Slack](https://github.com/LSSTDESC/descqa/blob/cd3e1e216fa5da76bb0dcf8d97a9267541c4fde4/run_master.sh#L10) and https://github.com/LSSTDESC/descqa/blob/cd3e1e216fa5da76bb0dcf8d97a9267541c4fde4/run_master.sh#L10

Would it be ok to just use the current default version of those or is it really better to specify a version and update this periodically?  I'd vote for making this generic.  I'll open a PR.

yymao:
This PR fixes #209 by loading the default modules (instead of a specific version).

### DIFF
--- a/run_master.sh
+++ b/run_master.sh
@@ -7,8 +7,8 @@
 set -e
 
 # load needed modules
-module load gsl/2.1
-module load cray-fftw/3.3.6.2
+module load gsl
+module load cray-fftw
 
 # activate python env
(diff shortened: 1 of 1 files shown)


heather999 (approved):
Looks Good!

PR #221, opened 2021-04-08
Use `setup_old_python.sh`

yymao:
DESCQA directly uses the python executable. For it to work with the new desc-python, DESCQA needs to be run within that shifter image. That wasn't expected in the original design.

For now I just switch to `setup_old_python.sh` in `run_master.sh` so that things should continue to work. We still need to find a long-term solution.

### DIFF
--- a/run_master.sh
+++ b/run_master.sh
@@ -7,7 +7,7 @@
 set -e
 
 # activate DESC python environment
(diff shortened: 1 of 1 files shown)


evevkovacs (approved):

yymao:
Thanks, Eve. Long-term solution still needed. See #222.

PR #227, opened 2021-11-11
Revert to using current python

### LINKED ISSUE #226: Revert to using the desc-python env at NERSC
Reverting to use desc-python at NERSC rather than desc-python-old.  We had made this switch when desc-python was a shifter image, but now desc-python is again natively installed.

yymao:
This PR reverts the change we made in 4c46f34 and fixes #226

### DIFF
--- a/run_master.sh
+++ b/run_master.sh
@@ -7,7 +7,7 @@
 set -e
 
 # activate DESC python environment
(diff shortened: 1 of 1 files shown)


heather999 (approved):
Looks good!

PR #228, opened 2021-11-11
point to desc-python rather than desc-python-old

heather999:
Addresses #226 
Now that `desc-python` is again natively installed at NERSC, and `desc-python-old` is pointing to a shifter image, descqa needs to use `desc-python` as it was doing originally.


yymao:
This is a duplicate of #227

heather999:
sorry - closing!

PR #235, opened 2023-02-05
SRV first pull request

patricialarsen:
Adding the SRV branch to DESCQA, this has additional tests related to the validation of Rubin data. Note that it relies on a development version of GCR-catalogs and GCR which are not yet merged as of 2/5, but are available in the forked versions at the patricialarsen repositories. In future updates these will be merged into the LSSTDESC repositories


patricialarsen:
@yymao added you as a reviewer to confirm if you're happy for me to push this into a separate branch and then refine from there. There remain a few issues to address such as the hard-coding of paths. I can also fix all these before pushing if you like. 

There also is a note on this pull request that the GCR and GCR-catalogs repositories would also require updates, they should be backwards compatible so I could merge them directly but would you like these in separate branches to avoid requiring an MPI4py dependency?

yymao:
Thanks @patricialarsen.  Yes, merging this to a branch here is good. 

Re: gcr/gcr-catalogs. Those can be PR to the main branches but can we make the mpi4py dependency optional? The code can check if mpi4py exists, and raise errors only if it is needed but not installed?

